PR #12, opened 2025-08-10
Add SpEL evaluation endpoint and implement secure bean access logic

### LINKED ISSUE #14: Spel evaulation
We need to be able to handle custom SPEL evaluation.

What it means is that the user submits a SPEL expression from the front-end, like this:

```
#{@myBean.isFeatureFlagOn() ? "1" : "0"}
```

And what needs to be done is that we need to be able to evaluate this SPEL expression inside the running Spring Boot application and provide the result to the end user.

We need a new actuator endpoint for that as well.

NikitaKirilloff:


mipo256 (changes requested):
Although this is a good start, I think that the current implementation needs to be significantly revisited.

P.S: What is also missing in current implementation - we need to add an `Authority` for this operation.

mipo256 (changes requested):
We're on the right track generally, a couple of comments

PR #31, opened 2025-08-20
GH-26 Add flat classpath discovery for JAR metadata

### LINKED ISSUE #26: Implement the ClassPathDiscoverer
Implement the `ClassPathDiscoverer` interface. An implementation should find all the entries of the classpath of the running application. In 99% of cases, that would be the bunch of jars in the Spring Boot fat JAR.

NikitaKirilloff:
- Introduce FlatJarClassPathEntry for simple JAR dependency representation
- Add FlatMavenMetadataExtractor and FlatGradleMetadataExtractor
- Create FlatClassPathDiscoverer that uses all available JarMetadataExtractor beans
- Add Spring Boot auto-configuration for flat classpath discovery


mipo256 (changes requested):
We have a decent draft, but we have something to work upon.

mipo256:
We do not need it for now.

PR #68, opened 2025-09-09
GH-48 Provide GraalVM native image build for master

### LINKED ISSUE #48: Provide GraalVM image build for master
What I think can be a good killer-feature of our solution is a native image build for master.  The idea is that the Axile Master will potentially be  quite a large service by its own, and therefore it may potentially consume a lot of resources on the clients's side. We may want to provide a native image of the master in order to reduce the resources consumed substantially.

**Problems:**

It is possible that we would allow the users to extend the master capabilities similar to what SBA does (providing custom implementations of interfaces etc). In this case, the way we provide native image build for Master is not clear at all. We need to think about it.

NikitaKirilloff:
- Added new GitHub Actions workflow that builds the native image of the `master` module using Liberica NIK.
- Disabled PMD and ErrorProne for AOT/native-related tasks to avoid false failures during
- New GitHub Actions pipeline ensures master native image can be built successfully.
- Verified basic master APIs after build: /beans, /info, /env.


NikitaKirilloff (commented):
@mipo256 please check

PR #178, opened 2025-10-07
GH-163 Added logic for changing environment property

### LINKED ISSUE #163: Integrate mutation of properties capabilities into the UI
We need to integrate the ability to change properties in the environment into the front-end.

Recently, the backend [got a new endpoint](http://158.160.200.59/api/axile/swagger-ui/index.html#/Property%20Management%20API/changePropertyValue) for mutating the given property on the given application instance. We need to think about the ways of integrating it into the front-end.

I can think of a new component (maybe also from `antd` design) that should be displayed right after the property value, and on click the user should be able to provide an input string, that is going to be later used as the new property value.

We also have an edge case of some sort. We need to be able to differentiate between the `null` as null pointer, and `"null"` as string.

AshotSargsyan1:


mipo256 (changes requested):
Very good graft, still, we have some work do

PR #181, opened 2025-10-08
GH-159: migrate to YAML OpenAPI spec from Annotations (Beans, Login, Wallboard) (#159)

### LINKED ISSUE #159: Migrate to YAML OpenAPI spec from Annotations
Currently, our master has the API that is covered with openapi swagger annotations such as `@Operation`, `@Parameter`, `@ApiResponse` and others. The end goal of having these annotation is to have auto-generated swagger UI so that the contract of backend is clear for front-end.

There are couple of problems with this approach though:

- First problem is that these annotations make the code of the API layer difficult to read. The code is drowning in these documentation-related annotations
- If we would want to provide our master API spec to third party for any reasons, it would not be possible to do that easily.
- It is possible that the documentation via swagger annotations does not exactly match the reality (see #157 as an example)

So, I think, we need to explore the following approach: 

- We might want to remove all the swagger annotations from the API layer.
- At the same time, create the `openapi.yaml` file that is going to define our API contract layer.
- We also want to have auto-generated models from the API layer (request/response models)
- Remove the hand-written request/response models that we have earlier

sergeycherkasovv:


NikitaKirilloff (changes requested):
Also consider splitting API description files if possible.

And please check Swagger before and after transferring the API description logic to the YAML file

PR #225, opened 2025-10-16
GH-218 Revisited components structure. Removed redux from most components

### LINKED ISSUE #197: Revisit the Redux Slices design
Currently, the Redux slices in the project have inconsistent structure which can be simplified. Moreover, it is also the common case that we do not really need Redux in a majority of components.

Thus, we need to improve the abstraction to make them more maintainable and easier to extend.

### LINKED ISSUE #199: Remove the searchText and filteredEntities from redux slices
After the online discussion it appeared that we do not really need to store the searchText and the filtered entities in the redux slices. Moreover, their existence produce an obstacle in implementation of the #197 issue.

So, this ticket needs to be dealt with prior to #197

### LINKED ISSUE #217: Caches: Reload animation triggers on all reload icons instead of the clicked one
When clicking on a specific reload icon in the ui interface (Caches), the loading animation is applied to all reload icons simultaneously. This creates a confusing user experience as it's unclear which specific element is being reloaded.

Current Behavior:
1. User clicks any reload icon
2. All reload icons in the view start spinning animation
3. User cannot identify which specific section/component is loading

Expected Behavior:
1. User clicks a specific reload icon 
2. Only that clicked icon shows loading animation 
3. Other reload icons remain in their default state

**Additional Consideration:
It might be worth reevaluating the success notification approach for cache clearing operations. Currently, all successful operations show a generic "Cleared" message. We could explore making these notifications more specific by including the cache name (e.g., "Cleared vets", "Cleared pets") to provide better user feedback.** - WDYT? @mipo256

### LINKED ISSUE #218: Move all requests from RTK to the components
Move all requests from RTK slices to component-level.
Create a hook responsible for fetching data.

AshotSargsyan1:


mipo256 (approved):
LGTM

PR #241, opened 2025-10-21
GH-200 added custom actuator endpoint ( id="axile-env" ), to Clarify the "Winning" (primary) Property Source

### LINKED ISSUE #200: Improve /actuator/env Endpoint to Clarify the "Winning" Property Source
Currently, the /actuator/env endpoint returns a comprehensive list of all PropertySources and their properties. While this is useful for seeing all available configuration, it can be difficult to determine which specific property value is actually being used by the application (the "winning" property) when the same key is defined in multiple sources (e.g., application.yml, environment variables, command-line arguments, etc.).

The Spring framework resolves properties based on the order of PropertySources, where a source higher in the list overrides ones below it. The current actuator output shows all values but doesn't explicitly mark the one that takes precedence.

Proposed Solution
We should enhance the endpoint's response to clearly indicate the effective, or "winning," property for each key.

NikitaKirilloff:


mipo256 (changes requested):
Solid draft, still have something to work on

PR #242, opened 2025-10-21
GH-236 Created 'Conditions' page with real data

### LINKED ISSUE #236: Create  'Conditions' page
Create a 'Conditions' page with real data.

AshotSargsyan1:


mipo256 (changes requested):
I think we're heading in the good direction, just a couple of comments to resolve

PR #259, opened 2025-10-25
GH-258: moved versions of common dependencies to the common module (#258)

### LINKED ISSUE #258: Move versions of common dependencies to the `common` module
The `jsonwebtoken` library is used in many **modules** of our project. To avoid potential version mismatch issues, it was decided to move the library version to the **common** module.

We have some plans to use the `Instancio` library, so it was decided to move it to the **common** module.

sergeycherkasovv:


NikitaKirilloff (approved):
made some minor changes

PR #261, opened 2025-10-25
GH-260: Fixed the CyclomaticComplexity error reported by PMD (#260)

### LINKED ISSUE #260: Find out why PMD is reporting a `CyclomaticComplexity` error.

sergeycherkasovv:
While working on task #260, we found that **PMD** counts the `throw` statement when calculating method complexity. I did not find this case [documented](https://docs.pmd-code.org/latest/pmd_rules_apex_design.html#cyclomaticcomplexity), though I may not have searched thoroughly.

Based on this, it was decided to move the **null-check** logic into a separate method `checkEmptyUserAndUserName` to avoid the **CyclomaticComplexity** error.


NikitaKirilloff (approved):
That's right, well done @sergeycherkasovv. Each throw actually increases the cyclomatic complexity.

PR #313, opened 2025-11-07
GH-289: parallel build test (#289)

### LINKED ISSUE #289: Linked issue job runs more often than required
Our Linked issue check job (introduced by #271) is a bit flawed now. For instance:

1. It runs on every master commit, which is wrong
2. It runs twice in case both the front-end and the backend was affect by some change, which should not be the case.

sergeycherkasovv:


mipo256:
I've cherry-picked [your commit manually and revisited an implementation a bit](https://github.com/axelixlabs/axelix/commit/920dd36d0db3bdc87d2a61afd49ec6ba003cc626). I made it simpler just by introducing a separate workflow for the linked issues check. I think this way it would be much more intuitive.

The authorship is preserved.

Thanks, mate @sergeycherkasovv !

PR #539, opened 2026-01-12
GH-507: proxy detection logic has been moved to the common module (#507)

### LINKED ISSUE #507: Unify proxy detection logic in the project
On the backend we have the spring proxy detection logic a bit spread across the project (e.g. the beans endpoint). We need to unify that. functionality.

sergeycherkasovv:


mipo256 (changes requested):
Looking at the PR, I am not really sure that we should introduce these changes in a way they currently are. 

On one hand, it is a good idea to unify proxy detection logic, on the other hand, introducing the dependency to `common:api` in `common:utils` is not a good idea, I think, since the API module is supposed to be consumed only by `master` and `sbs`.

Furthermore, introducing a dependency for `spring-web` in common is also not a desirable thing to be doing since that means that we introduce spring in every module that includes `common:utils`.

We need to give these changes a bit more thought.

mipo256:
We're going to return to that question if we see fit in the future, right now, it does not seem to be worth it

PR #768, opened 2026-02-25
GH-754 Optimized loggers page

### LINKED ISSUE #754: [Refactoring] Think about the ways the loggers DOM components sub-tree can be optimized
The loggers component with tabs is quite a mess right now. We need to brainstorm and figure out a ways to simplify it, and apart from that, we want it to perform better, since now every tiny change requires the re-rendeing of the entire components tree, including both tabs.

AshotSargsyan1:
Closes GH-754


mipo256 (approved):

mipo256:
Great work, @AshotSargsyan1! Merged.

PR #781, opened 2026-02-27
Feature/gh 778

### LINKED ISSUE #778: [Feature] Create a GitHub Action to deploy the website on a VM
Create a GitHub Action to deploy the website on a VM.

AshotSargsyan1:
Closes GH-778

### DIFF
--- a/.github/workflows/website_deploy.yaml
+++ b/.github/workflows/website_deploy.yaml
@@ -2,13 +2,13 @@ name: Deploy Website
 
 on:
   push:
-    branches: [feature/GH-778]
-    # paths:
-    #   - 'website/**'
+    branches: [master]
+    paths:
+      - 'website/**'
 
 jobs:
   deploy:
-    runs-on: ubuntu-latest 
+    runs-on: ubuntu-latest
 
     defaults:
       run:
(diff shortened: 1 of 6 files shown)


mipo256 (changes requested):
1. The container process should not be launched by `root` user
2. We should both deploy nginx & nextjs via docker compose, and we need to cofnigure nginx to be able to server https traffic

.github/workflows/website_deploy.yaml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Build Docker image on server and restart container
+        uses: appleboy/ssh-action@v1.2.5
+        with:
+          host: ${{ secrets.WEBSITE_PRODUCTION_VM_HOST }}
+          username: ${{ secrets.WEBSITE_PRODUCTION_VM_USER }}
+          key: ${{ secrets.WEBSITE_PRODUCTION_VM_PRIVATE_KEY }}
+          script: |
+            set -eux
+            cd /home/${{ secrets.WEBSITE_PRODUCTION_VM_USER }}/axelix-website
+            sudo docker compose -p axelix-website down || true
```
mipo256: I think that is fine for now, we may in the future explore the docker rootless mode, just for secrutiy purposes. But that however would require some IaaC for VMs in the cloud, which we do not have right now

mipo256 (approved):
We're generally fine, let's just change the `latest` for the conrecete sha for nginx, and we're good to go

Issue #790, opened 2026-03-02
[Feature] Migrate Master to Java 25


mipo256:
Master currently runs on Java 25. We can update it to run on Java 25 (that _seemingly_ should not force us to migrate master to Spring Boot 4, so we're good here).

So the master gradle build file should be adjusted, along with the `Dockerfile` for building the Axelix Master.

Chrimle:
Looks like it's currently a mix between [JDK17](https://github.com/axelixlabs/axelix/blob/master/.github/workflows/release.yaml#L38) and [JDK21](https://github.com/axelixlabs/axelix/blob/master/.github/actions/build-backend/action.yaml#L16)?

Issue #84, opened 2025-09-19
Migrate to Gradle 9


mipo256:
Migrate the build to use Gradle 9.

mipo256:
Implemented with #796

PR #796, opened 2026-03-04
GH-790: Migrate Master to Java 25

CoolXaleX:
Bumps the Java toolchain in master/build.gradle.kts from 21 to 25, and updates the release pipeline to set up JDK 25

Seems like that's enough.


build.gradle.kts
```diff
@@ -36,14 +37,18 @@ subprojects {
     }
 
     plugins.withType<JavaPlugin> {
+        dependencies {
```
mipo256: This is a bit obscure...

Why would we need to include only for the test runtime a specific version of the JUnit platform launcher? What problem are we solving with that?

mipo256:
Related #84

PR #826, opened 2026-03-15
GH-825: Moved the custom source descriptions the `Env API` to the `starter-domain` module

### LINKED ISSUE #825: [Refactoring] Move the Environment API logic from `axelix-spring-boot-3` to the `starter-domain` module

sergeycherkasovv:


mipo256 (commented):
I do not think we need to merge this PR, and probably should not explore it for a couple of reasons.

1. This PR adds even more code than existed prior to that change
2. The property sources might be different in different versions of spring boot/cloud, and they may have different meanings.

I think there is no need to introduce it, at least for now.

PR #828, opened 2026-03-16
GH-827: Moved the configuration properties setup logic to the `starter-domain` module

### LINKED ISSUE #827: [Refactoring] Move the configuration properties setup logic to the `starter-domain` module

sergeycherkasovv:


mipo256 (changes requested):
I think we're headed in the right direction. I understand why certain actions were made, like validation decoupling from data creation, but I think we need a better solution. I'll think about it myself.

Issue #829, opened 2026-03-17
[Refactoring] Move the ThreadDump API logic to the `starter-domain` module


sergeycherkasovv:

mipo256:
See the related PR

PR #830, opened 2026-03-17
GH-829: Moved the ThreadDump API logic to the `starter-domain` module

sergeycherkasovv:


mipo256 (commented):
I do not think we need to merge it.

The general guideline is that less code is better. Here, we can of course extract common part to `starter-domain`, but this extraction is not justified (at least now), since by extracting couple of lines of code, we create new abstractions, and we ultimately increase the numebr of LOC.

PR #834, opened 2026-03-17
GH-811 Changed how metric trends are displayed on the chart

### LINKED ISSUE #811: [Feature] Change how metric trends are displayed on the chart
Currently, the chart displaying the dynamics of the selected metric on the Metrics page is visually overloaded, which makes it harder to read.

Image

It is proposed to implement something similar to the example below (see the image), where a marker is displayed only at the end of the line. This should improve the readability of the chart. The chart should contain a single green line.

Image

**P.S.**
Please consider the possibility of disabling the visual display of the markers currently shown along the chart line, rather than removing them completely. The tooltip with metric information should still appear when hovering over the corresponding area of the line, as long as an invisible marker exists there.

AshotSargsyan1:
Closes GH-811


mipo256 (approved):
lgtm

PR #837, opened 2026-03-18
GH-836: Added test for MCP API

### LINKED ISSUE #836: [Feature] Add test for MCP API

sergeycherkasovv:


master/src/test/java/com/axelixlabs/axelix/master/api/McpToolApiTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Test
+    void should() {
+        // when.
+        ResponseEntity<McpToolFeedResponse> response = restTemplate
+                .withoutAuthorities()
+                .getForEntity("/api/external/mcp/tools-feed", McpToolFeedResponse.class);
+
+        // then.
+        assertThat(Objects.requireNonNull(response.getBody()).tools())
```
mipo256: 1. Let's use jsonunit here as we often do when we test the API layer.
2. I think, the best way of writing the test here would be to check if all of the tools in a way that we expect are present in the response (ignoring the description, it's just too damn large). Something like

```(java)
    assertThatJson(jsonResponse)
        .node("tools") // Specify the path to the array
        .isArray()
        .containsOnly(
            value(
                """
                   { "title" : "My title", "description" : "#{json-unit.ignore}" ... }
                """
            ),
            // other values here
        );
```

mipo256 (changes requested):
We're heading in right direction, still so work is required.

PR #846, opened 2026-03-20
GH-840: Added Cache-related MCP tools

### LINKED ISSUE #840: [Feature] Add Cache-related MCP tools
We need to add an MCP endpoints related to caching, i.e.

1. Clear all the caches
2. Clear caches inside the given cache manager
3. Clear specific cache within cache manager
4. Clear specific entry by key within cache 

Basically, this API is similar to what we allow in the UI, we just need to add this capabilities as MCP Tools.

sergeycherkasovv:


mipo256 (changes requested):
Okay, so, in general, look good.

The tests need to be revisited a bit still. I'll do that myself.

PR #847, opened 2026-03-20
GH-842 Fixed a bug related to MCP tool cards heights

### LINKED ISSUE #842: [BUG] Incorrect height calculation for multi-row MCP Tools page
We still have certain problems with the MCP Tool cards. For instance, in the current grid layout we have rows as subgrids, and these rows inherit their height from the row that is the highest. That is correct. But that breaks because in case we have multiple rows, that all of a sudden breakes.
 
Image

AshotSargsyan1:
Closes GH-842


mipo256 (commented):
So, as a result of a discussion with @AshotSargsyan1, we decided to fall back to the simple hardcode of the description row height for now.

The problem is that the suggested solution, although it really works, brings a lot of complexity, which does not seem to be justified from the perspective of software maintenance. It seems that it would be easier for us to just change the hardcoded height if needed, rather than maintain this code.

See b9da78c9

Issue #870, opened 2026-03-23
[Feature] Implement AuthorityResolver in starter domain


mipo256:
We need to provide an implementation for `AuthorityResolver` in `starter-domain`. The implementation should contain the mapping between endpoints (like `ActuatorEndpoint` --> `GlobalAuthority`). As a result, the `GlobalAuthority` enum itself must be revisited in accordance to the authorities stated in the #848. 

We can, probably, for the actual match re-use spring's `PathPatternParser`. In this case the actual matching function should be taken from outside (from starter modules).

Apart from that, we should try to get rid of the `PassthroughAuthorityResolver` usages in favor of a new implementation. This would require revisiting some tests.

PR #874, opened 2026-03-26
GH-870: Implemented AuthorityResolver in the starter-domain module

sergeycherkasovv:


sbs/starter-domain/src/test/java/com/axelixlabs/axelix/sbs/spring/core/auth/DefaultAuthorityResolverTest.java
mipo256: I think we can reduce the amount of code in this test class significantly if we use `@ParametrizedTest` an empty Optional, and when the single authority is resolved. Or maybe some other structure of `@ParametrizedTest` is applicable here. Please, explore this path @sergeycherkasovv

PR #876, opened 2026-03-28
GH-758 implement OAuth2 authentication for MCP server

### LINKED ISSUE #758: [Feature] We need to think of an authentication mechanism for the MCP server.

NikitaKirilloff:
closed #758 

## Summary by CodeRabbit

* **New Features**
  * Added OAuth2 Protected Resource Metadata endpoint for MCP integration.
  * Enhanced MCP support: centralized MCP endpoint identifiers, new MCP authorization/filtering, and support for Basic and Bearer authentication modes.
  * New MCP tools/endpoints exposed for feeds, wallboard, caches, and scheduled tasks.

* **Tests**
  * Expanded integration and unit tests covering OAuth2 metadata, MCP authorization flows, role extraction, and endpoint resolution.


master/src/main/java/com/axelixlabs/axelix/master/filter/McpBearerTokenFilter.java
```diff
@@ -0,0 +1,94 @@
...
+import static com.axelixlabs.axelix.master.autoconfiguration.web.WebAutoConfigurationConfiguration.API_EXTERNAL_PATH;
+
+/**
+ * Validates Bearer tokens for MCP endpoints.
+ *
+ * @author Nikita Kirillov
+ */
+@Component
+@ConditionalOnProperty(prefix = "axelix.master.auth.options.oauth2", name = "enabled", havingValue = "true")
+public class McpBearerTokenFilter extends OncePerRequestFilter {
```
NikitaKirilloff: Currently, this filter only performs token validation and restricts access to the MCP server via login and password.

common/utils/src/main/java/com/axelixlabs/axelix/common/utils/Lazy.java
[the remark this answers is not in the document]
mipo256: > I'm not sure the `@SuppressWarnings("NullAway")` annotation should be here. We have two constructors for the `Lazy` class, which means the `supplier` could be null. I think in this case it's better to add a check for null and throw a clear exception if it fails.

Yes, the `supplier` could indeed be null, but only in the case when `Lazy` is resolved. If the lazy is not resolved, then the supplier is a required parameter. So I think this suppression is fine.

WDYT? @NikitaKirilloff

master/src/main/java/com/axelixlabs/axelix/master/filter/McpAuthenticationFilter.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    protected void doFilterInternal(
+            @NonNull HttpServletRequest request,
+            @NonNull HttpServletResponse response,
+            @NonNull FilterChain filterChain)
+            throws IOException {
+
+        String authHeader = request.getHeader("Authorization");
```
mipo256:
```suggestion
        String authHeader = request.getHeader(HttpHeaders.AUTHORIZATION);
```

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/McpOAuth2MetadataController.java
```diff
@@ -0,0 +1,57 @@
...
+@ExternalApiRestController
+@ConditionalOnProperty(prefix = "axelix.master.auth.options.oauth2", name = "enabled", havingValue = "true")
+public class McpOAuth2MetadataController {
+
+    private final String issuerUri;
+    private final String mcpServerFullPath;
+    private final String scopes;
+
+    public McpOAuth2MetadataController(OAuth2Properties oAuth2Properties) {
+        this.mcpServerFullPath = oAuth2Properties.baseUrl() + WebAutoConfiguration.EXTERNAL_API_PATH + "/mcp";
```
mipo256: Isn't the endpoint servlet path suppose to be `spring.ai.mcp.server.streamable-http.mcp-endpoint`? In our case `/api/mcp`:

```(yaml)
spring:
  application:
    name: axelix
  ai:
    mcp:
      server:
        protocol: streamable
        name: "axelix-mcp-server"
        streamable-http:
          mcp-endpoint: "/api/mcp"
```

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
```diff
@@ -44,6 +48,7 @@ public record OAuth2Properties(
         String issuerUri,
         String clientId,
         String clientSecret,
+        String baseUrl,
         String redirectUri,
         @Nullable String usernameClaim,
         String scopes) {
```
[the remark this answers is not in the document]
mipo256: We're gonna revisit it from top to bottom in another dedicated task

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
```diff
@@ -44,6 +48,7 @@ public record OAuth2Properties(
         String issuerUri,
         String clientId,
         String clientSecret,
+        String baseUrl,
         String redirectUri,
```
mipo256: We we're still accepting the `redirectUri` as a configuration property? This one is an artificially constructed property. Let's remove it from record components and create a separate property accessor:

```
    public String redirectUri() {
        return baseUrl + WebAutoConfiguration.EXTERNAL_API_PATH
                + ApiPaths.OAuth2Api.CALLBACK;
    }
```

master/src/main/java/com/axelixlabs/axelix/master/filter/McpAuthenticationFilter.java
mipo256, line 1: We have to do weird stuff inside this filter, like inject `ObjectProvider` and do `if` checks for null just because this filter may work with Bearer tokens and may work with basic auth. Why not just create two separate filters for these different scenarios? I guess we should

master/src/main/java/com/axelixlabs/axelix/master/filter/McpAuthenticationFilter.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        } catch (Exception e) {
+            log.debug("Basic auth validation failed: {}", e.getMessage());
+            sendForbidden(response, "Invalid username or password");
+        }
+    }
+
+    private void sendUnauthorized(HttpServletResponse response) {
+        response.setStatus(HttpServletResponse.SC_UNAUTHORIZED);
+        response.setHeader("WWW-Authenticate", "Bearer resource_metadata=\"" + resourceMetadata + "\"");
```
mipo256: That's not so easy. Why are we sending the auth schema as `Bearer` every time? Even if the end users would work with `Basic` auth schema. Hardcoding it here is not going to work.

mipo256 (changes requested):
I think we're on the right track. Still, there are a couple of issues to consider.

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/McpOAuth2MetadataController.java
NikitaKirilloff: Yes, exactly, thx

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
NikitaKirilloff: Yes, I forgot to delete it, thx

mipo256 (approved):

PR #887, opened 2026-04-01
GH-886 Allow for creation of extensions to the Front-end codebase

### LINKED ISSUE #886: [Feature] Prepare Community version for Enterprise adaptation
We need to prepare the current Community version of our product so that it can be easily extended and adapted for an Enterprise version.

AshotSargsyan1:
Closes GH-886

## Summary by CodeRabbit

* **New Features**
  * Enterprise edition support with dynamic enterprise-module loading
  * Simple extension registry for registering and using custom components
  * Added a new CodeSnippetArrowIcon for UI

* **Bug Fixes**
  * Prevented unintended click event propagation in annotations popover

* **Chores**
  * Added enterprise submodule and updated build/TypeScript/ESLint config to support enterprise code and headers

[Review Change Stack](https://app.coderabbit.ai/change-stack/axelixlabs/axelix/pull/887)


front-end/src/main.tsx
```diff
@@ -18,11 +18,25 @@
...
 import { createRoot } from "react-dom/client";
 
+import { EEdition } from "models";
+
 import { App } from "./App";
 
 import "./index.css";
 import "./customizedTable.css";
 
+if (import.meta.env.VITE_EDITION === EEdition.ENTERPRISE) {
```
mipo256: I'd move the `EEdition` enum right into this file, since `EEdition` is not a regular enum that can be used anywhere in the source code. 

Even more, I am not sure that we need an enum, I think we can get by with just a constant

```
const ENTERPRISE = "enterprise";
```

mipo256 (changes requested):
So, great. I think we're moving where we should. At the same time, I have a few minor picks, please check them out.

More general questions include the following:

1. We're not making the enterprise edition the submodule of the community edition. If we're not, then how `enterprise` the folder supposed to be managed? Manually cloning, I think, is a bit error-prone.

2. Apart from that, I start to notice the unhealthy pattern -  we kind of starting to leak the enterprise "stuff" into community version. Like the `i18n` json dictionary, now it contains both the enterprise keywords and community ones. I think in the community we must have a way either 

a. To extend the dictionary dynamically
b. Or, what can even be better, we can instruct the `react-i18next` to look into two dictionaries - first , if found, in the enterprise dictionary, and then in the community.

Same goes for SVG icons. I mean, in general, if we're going to start to mix up icons used in community edition with enterprise edition, then it would lead to problems, such as unsued icons, someone may delete them, or someone in community may reuse the image in enterprise.

The problem is that the files in the community version are licensed under LGPL. And the enterprise edition would have the other, proprietary license. So mixing this up is just not an option, unfortunately.

front-end/src/main.tsx
```diff
@@ -23,6 +23,20 @@ import { App } from "./App";
 import "./index.css";
 import "./customizedTable.css";
 
+const ENTERPRISE = "enterprise";
```
mipo256, line 26: We're going to build the unified distribution. I would just count on the presence of the `./enterprise/src/index.ts` realtive to the current file.

front-end/.gitignore
```diff
@@ -13,6 +13,7 @@ dist-ssr
 *.local
 .env
 .editorconfig
+enterprise
```
mipo256: If we're gonna ad this as the git submodule we're not gonna need it.

mipo256 (changes requested):
The problem is deeper. 

Since we're going to attempt to provide a unified distribution, we need to be able to build it all together, but load the extensions only if the license key is available.

This problem is a bit more complex. Let us research the way the license key is going to be provided, and then we're going to return back to this.

front-end/.gitignore
AshotSargsyan1: This was originally done for safety reasons and has already been removed.

PR #896, opened 2026-04-03
Polishing: provided a junit test template for negative auth cases

mipo256:


master/src/main/java/com/axelixlabs/axelix/master/filter/CookieBasedJwtAuthorizationFilter.java
NikitaKirilloff:
```suggestion
*/
```
I think we can delete `@SuppressWarnings("NullAway")`

PR #905, opened 2026-04-07
GH-839 Designed transaction inspection capabilities

### LINKED ISSUE #839: [Feature] Design of Transactions Inspection capabilities
This ticket outlines the initial version of the deep transaction inspection UI design as we see it.

**Goals:**

The primary goal is to communicate to the end user the information about the particular transaction executions in the progressive disclosure UX/UI design pattern. In particular, we want to aim the end user in 

1. Discovering the "N + 1" SELECT problems.
2. Discovering the SQL queries that took the longest inside the transaction.
3. Discovering the potential non-database interactions inside the transaction (i.e. potential network calls or something) 

The end-user should check the transaction control page for the problem of having long-lasting transactions (or suspicion of having long-lasting transactions) and find out the reasons why they are long-lived.

**Overall design:**

1. Once the user clicks on the transaction record dot on the transactional chart, we have the accordion opened underneath the transaction statistics

2. The end-user should have the ability to close the accordion by clicking on the cross mark at the right side of the created accordion

3. The accordion should be broken down into two sections:
- Metadata section on the left, with fields: "Start Time", "End Time" and "Queries Count"
- The bar chart: This bar-chart has a fixed size for each transaction inside any feed. The single bar that is broken down into parts:

Image

Each "red" part is supposed to represent an individual query inside the transaction. When hovering over the query, we should show a tooltip, similar to what we have now on the transactions chart:

Image

So we should have the "Start Time" and the "Duration" of the query in the tooltip. The styling of the tooltip is presumably the same.

4. The end-user should be able to click on each individual query inside the transaction, and once he does, we need to open a popover on either the left or the right side. The exact side where the popover should be opened must be determined by the JavaScript in the browser, which needs to check the available space on both sides and select the one that has the largest space available. 

The popover itself should contain the actual, formatted SQL. If there is not enough space, either vertical or horizontal (which is not very likely but can happen in real-world applications), we should create the scroller. On the top of the Popever, we should display the "Total Time: X"—the total duration of the given query execution.

AshotSargsyan1:
Closes GH-839


front-end/src/pages/Transactional/TransactionalList/TransactionalAccordionBody/ExecutionDetails/QueriesTimeline/QueryBar/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+            setSelectedQuery(null);
+            setTimeout(() => setIsClosing(false), 300);
+        }
+    };
+
+    return (
+        <Popover
+            trigger="hover"
+            content={<QueryPreview query={query} />}
+            open={isQuerySelected || isClosing ? false : undefined}
```
mipo256: I'm not 100% sure, but it seems that we do not need the `open` property at all here. 

The `Popover` should just be triggered by hovering on it. Making it open in case the current query is selected or when the popover with the SQL query is closing, I think, is not a good idea, I think.

front-end/src/pages/Transactional/TransactionalList/TransactionalAccordionBody/ExecutionDetails/QueriesTimeline/styles.module.css
```diff
@@ -0,0 +1,19 @@
...
+    background-color: var(--axelix-light-grey);
+    border: 1px solid var(--axelix-medium-grey);
+    margin-bottom: 8px;
+}
+
+.Axis {
+    display: flex;
+    justify-content: space-between;
+    color: var(--axelix-bold-grey);
+    font-family: monospace;
```
mipo256, line 18: Do we really want to override the `font-family` for the Axis? I thought that using the same consistent font across all the components is the standard practise.

mipo256 (approved):
I have made some minor changes, please take a look @AshotSargsyan1.

Apart from that, the draft is very promising, great work!

front-end/src/pages/Transactional/TransactionalList/TransactionalAccordionBody/ExecutionDetails/QueriesTimeline/styles.module.css
AshotSargsyan1: Yes, this is standard practice, but I think there can be exceptions. In this case, it simply gives a more technical look. Overall, it's fine to override things.

PR #906, opened 2026-04-07
GH-873 Most of the service access control implementation is done

### LINKED ISSUE #873: [Feature] Add service access control
Add service access control.

AshotSargsyan1:
Closes GH-873


front-end/src/helpers/globals.ts
```diff
@@ -123,3 +124,26 @@ export const createWallboardFilterSearchParam = (
 ): string => {
     return `${key}:${operator}:${operand}`;
 };
+
+export const getCookie = (name: string): string | undefined => {
```
mipo256: 1. We need tests for these two functions
2. I think we might also make them a bit better

front-end/src/pages/GarbageCollector/GCFirstSection/index.tsx
```diff
@@ -106,6 +109,7 @@ export const GCFirstSection = ({ loadGCStatus, isLoggingStatusEnabled }: IProps)
                                     loading={downloadFileLoading}
                                     onClick={downloadFileHandler}
                                     className={styles.ActionButton}
+                                    disabled={!GCAccess}
```
mipo256: That should not be guarded

mipo256 (commented):
Thats generally good, some minor issues, I'll polish them myself

PR #909, opened 2026-04-08
GH-891: Revisited authorities in tests

### LINKED ISSUE #891: [Feature] Revisit authorities in tests
In all the tests the `GlobalAuthority` needs to be changed to a `DefaultAuthority`. We do not need the `GlobalAuthority` as such.

It is not as easy as it sounds since `DefaultAuthority` enum has a different semantic meaning from `GlobalAuthority`. This is because `GlobalAuthority` is a per-endpoint basis (which is not really what we want, at least for now). So all of the tests needs to be adhusted accordingly, i.e. `DefaultJwtDecoderTest`/`DefaultJwtEncoderTest` and others.

sergeycherkasovv:


common/auth/src/main/java/com/axelixlabs/axelix/common/auth/core/jwt/DefaultJwtEncoderService.java
```diff
@@ -95,15 +95,15 @@ private void validateUser(User user) {
     }
 
     private List<JwtRole> buildRoleClaims(User user) {
-        return user.getRoles().stream().map(this::toJwtRole).toList();
+        return user.getRoles().stream().map(this::toJwtRole).collect(Collectors.toList());
```
sergeycherkasovv, line 100: Java 11 does not support this syntax)

common/auth/src/test/java/com/axelixlabs/axelix/common/auth/DefaultJwtDecoderServiceTest.java
```diff
@@ -45,158 +52,231 @@
...
 
+    @Test
+    void shouldThrowOnExpiredToken_TokenWithNullNameRoles() {
+        Role role = new DefaultRole(null, Set.of(DefaultAuthority.CACHES_CLEAR));
+        User user = new DefaultUser(USER_NAME, PASSWORD, Set.of(role));
+        String token = jwtEncoderService.generateToken(user);
+
+        assertThatThrownBy(() -> jwtDecoderService.decodeTokenToUser(token)).isInstanceOf(JwtParsingException.class);
+    }
+
```
sergeycherkasovv: This test shows that `DefaultJwtEncoderService` and `DefaultJwtDecoderService` behave inconsistently, because. `DefaultJwtEncoderService` handles null `Role` values fine, but `DefaultJwtDecoderService` throws an exception in this scenario. This raises the question, should `DefaultJwtEncoderService` also throw an exception?

I believe we should throw an exception if the role has no **name**.

common/auth/src/test/java/com/axelixlabs/axelix/common/auth/service/DefaultJwtDecoderServiceTest.java
```diff
@@ -45,158 +52,231 @@
...
+                        .filter(role -> role.getName().equals("VIEWER"))
+                        .findFirst()
+                        .orElseThrow()
+                        .getAuthorities())
+                .isEmpty();
+    }
 
-        assertThat(readRole.getAuthorities()).contains(GlobalAuthority.BEANS);
+    @Test
+    void shouldDecodeValidJwtToken_MultipleRoles() {
```
sergeycherkasovv, line 99: This test concerns me since we rarely use **multiple roles** in practice, but we still support it. Do we plan to keep **multiple roles** support?

Same for `components`, I haven't written tests for their population/correctness. Do we plan "role-in-role" behavior?

sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/JwtAuthAutoConfiguration.java
```diff
@@ -70,7 +70,7 @@ public JwtDecoderService jwtDecoderService(AuthProperties configurationPropertie
     public AuthorityResolver authorityResolver() {
         return new DefaultAuthorityResolver((pathTemplate, actualPath) -> {
             PathPattern parse = new PathPatternParser().parse(pathTemplate);
-            return parse.matchAndExtract(PathContainer.parsePath(actualPath)) != null;
+            return parse.matchAndExtract(PathContainer.parsePath("/actuator" + actualPath)) != null;
```
sergeycherkasovv, line 83: During a call with @NikitaKirilloff , we decided to add the `/actuator` prefix at this stage and for this task so that the **resolver** works correctly.)

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/auth/JwtAuthorizationFilterTest.java
```diff
@@ -76,99 +108,228 @@
...
+            "axelix.prop.test.name=axelix-beans",
+        })
+@Import({
+    JwtAuthorizationFilterTest.JwtAuthorizationFilterTestConfiguration.class,
+    AxelixCachesEndpoint.class,
+    DefaultCacheOperationsDispatcher.class,
+    EnvironmentTestConfig.class
+})
+@DirtiesContext(classMode = DirtiesContext.ClassMode.BEFORE_EACH_TEST_METHOD)
 class JwtAuthorizationFilterTest {
```
sergeycherkasovv, line 126: When testing JwtAuthorizationFilter, I only used a few endpoints:
- `/actuator/axelix-caches/*` - accessible to **ADMIN**, **EDITOR**, and **VIEWER** roles
- `/actuator/axelix-beans` - accessible to all
- `/actuator/axelix-env` - **ADMIN** only

Ideally, we should test all endpoints, but I'm concerned the test would become too slow and large)

sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/JwtAuthAutoConfiguration.java
mipo256: I generally agree, thats okay

common/auth/src/test/java/com/axelixlabs/axelix/common/auth/service/DefaultJwtDecoderServiceTest.java
mipo256: > This test concerns me since we rarely use multiple roles in practice, but we still support it. Do we plan to keep multiple roles support?

Yeah, we need this. The abstraction assumes it. It it not that hard for us to support. Regarding the `components` - I agree. I do not think we need to cover it with tests yet. I think we might build something in enterprise edition, for now I think we should not touch it.

mipo256 (approved):
I have pushed some minor polishing changes.

Overall I think we're 90% good. However, I do not really think that the approach we took when testing the `JwtAuthorizationFilter` is the right one. 

There are quite a few infrastructural components in the system (Jakarta servlet filters are one of them) that do not really make much sense on their own. Writing unit tests for them is possible, but the setup would be quite excessive, and I do not think they will add much to the picture. 

For such infrastructural components I think we should prefer the approach that we took in the Axelix Master. We do not test the `CookieBasedJwtAuthorizationFilter` itself, rather, we test endpoints E2E. 

The goal of the `CookieBasedJwtAuthorizationFilter` (or any filter for that matter) is to intercept the request and do stuff. In our case we perform the access check and set the current user into the `SecurityContext `. I think that access check for each endpoint should be performed in the test class that is designed for each individual test endpoint. Like it is already done in the Axelix Master.

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/auth/JwtAuthorizationFilterTest.java
mipo256: I see that, and I think this approach is a bit wrong. The test would indeed grow in size, that is why I think we should consider somehow sharing the logic of `ProtectedEndpointExtension` with starter modules. I do not know exactly how for now, but we should probably leave it for another time. For now we have _some_ access checks, and I think we can be sure they are there.

common/auth/src/test/java/com/axelixlabs/axelix/common/auth/DefaultJwtDecoderServiceTest.java
mipo256: > This test shows that DefaultJwtEncoderService and DefaultJwtDecoderService behave inconsistently, because. DefaultJwtEncoderService handles null Role values fine, but DefaultJwtDecoderService throws an exception in this scenario

From the code I infer that this is the other way around. In any case, I do not think it is a big deal that one can handle `null` and the other do not. The `User` is in the package that is `@NullMarked`, so we would not have `null` roles.

PR #910, opened 2026-04-08
GH-902 Proxy token from starters to master

### LINKED ISSUE #902: [Feature] Proxy token from starters to master
Build a temporary token and send it in a request from starter --> master. Validate token on the master side. We can, I guess, create a separate filter for that, re-using the cookies-based seems to be a bit wrong here.

NikitaKirilloff:
close #902


sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/config/AuthProperties.java
```diff
@@ -51,6 +53,11 @@ public static class Jwt {
          */
         private String signingKey;
 
+        /**
+         * JWT token validity duration.
```
mipo256:
```suggestion
         * JWT token validity duration to be set when constructing the token.
```

sbs/axelix-spring-boot-3/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/JwtAuthAutoConfiguration.java
```diff
@@ -89,11 +100,12 @@ public IdentityAccessManager securityManager(
 
     @Bean
     public FilterRegistrationBean<JwtAuthorizationFilter> jwtAuthorizationFilterRegistration(
-            IdentityAccessManager identityAccessManager) {
+            IdentityAccessManager identityAccessManager, WebEndpointProperties webEndpointProperties) {
```
mipo256: I understand the rationale for this change, however I think we should roll it back.

Again, the problem is way broader. The problem is that how do we know what the actuator base path is? The `JwtAuthorizationFilter` is supposed to handle all the queries from Axelix Master, regardless of how the managed service was discovered. 

We have not yet solved this issue. We will do it separately. **The PR must be small**.

mipo256 (changes requested):
Good draft @NikitaKirilloff, but there are some issues still. 

1. Do not forget to transfer the tests for the source files along with the source files, like Jwt encoding facilities have been transfered to common so the tests should also.

2. There are a lot of small pitfalls, please, read all of them carefully and learn from them.

3. The overall algorithm is not really ideal. I left my thoughts on it.

master/src/test/java/com/axelixlabs/axelix/master/api/internal/SelfRegisteredApiTest.java
```diff
@@ -85,16 +89,16 @@ public class SelfRegisteredApiTest {
...
-                .asViewer()
+                .withRoleTokenInAuthorizationHeader(DefaultRole.SELF_REGISTRAR)
                 .postForEntity("/api/internal/service/register", defaultJsonEntity(JSON_REQUEST), Void.class);
 
         // then.
         assertThat(response.getStatusCode()).isNotNull().isEqualTo(HttpStatus.NO_CONTENT);
     }
 
     @ParameterizedTest
-    @EnumSource(InvalidAuthScenario.class)
```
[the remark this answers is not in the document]
NikitaKirilloff: ok

PR #915, opened 2026-04-09
GH-805 Implemented PostgreSQL and MySQL database support, integrated with Liquibase for schema migrations.

### LINKED ISSUE #805: [Feature] Incorporate support for databases other than SQLite
Incorporate the support for PostgreSQL and MySQL database alternatives. [JetBrains's recent report suggests](https://devecosystem-2025.jetbrains.com/tools-and-trends) that those are the most famously used, and from my empirical knowledge, I tend to agree here.

**TBD:**

1. Adjust the Auto Configurations if needed (Although I'm not sure this is going to be needed at all, since, ideally, we should be able to deduce the database in use from the JDBC URL alone).

2. Write tests with Test Containers for the APIs that interact with DB. Ideally, we can write our own JUnit extension to run the same tests across different databases. The logic inside the tests should not change, but the underlying database does.

### LINKED ISSUE #879: Think about the ways the SQLite database schema is going to evolve
We're soon going to have SQLite as the primary storage (a relational database for that matter, see #805). And we're going to have schema. Sure, for now, the data is ephemeral like the current Instances Grid. But later if we're going to store any state, like "real" state, e.g. some audit trails etc - we cannot just get away with it. We need to develop the migrations strategy.

NikitaKirilloff:
close #805 , #879


master/src/main/resources/db/changelog/mysql/db.changelog-mysql.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+            http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="mysql-001-create-instances-table" author="Nikita Kirillov" dbms="mysql">
+        <preConditions onFail="MARK_RAN" onError="MARK_RAN">
```
mipo256: I think halting is better here since if we encounter any error but we mark the migration as "ran", then we'll effectively have a mismatch between what schema is actually rolled out and the schema that the Axelix Master expects.

```suggestion
        <preConditions onFail="MARK_RAN" onError="HALT">
```

master/src/main/resources/db/changelog/mysql/db.changelog-mysql.xml
```diff
@@ -0,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+            http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <changeSet id="mysql-001-create-instances-table" author="Nikita Kirillov" dbms="mysql">
```
mipo256: The author of the migration would be put inside the database `databasechangelog` table. I do not really think that your first/last name would communicate anything meaningful to the users ("Axelix Team" would signal who's the author of the migration - the core Axelix project team). Apart from that, I do not think that having your name in all the databases backing Axelix is a good idea

```suggestion
    <changeSet id="mysql-001-create-instances-table" author="Axelix Team" dbms="mysql">
```

master/src/main/resources/db/changelog/mysql/db.changelog-mysql.xml
```diff
@@ -0,0 +1,61 @@
...
+            <not>
+                <tableExists tableName="instances"/>
+            </not>
+        </preConditions>
+
+        <createTable tableName="instances">
+            <column name="instance_id" type="VARCHAR(255)">
+                <constraints nullable="false" primaryKey="true" primaryKeyName="pk_instances"/>
+            </column>
+            <column name="name" type="TEXT">
```
mipo256: Let's fall back to `VARCHAR(N)` for all MySQL columns. We do not need `TEXT`, and `TEXT` on earlier versions of MySQL is going to cost us some overhead.

master/src/main/resources/db/changelog/mysql/db.changelog-mysql.xml
```diff
@@ -0,0 +1,61 @@
...
+            <column name="kotlin_version" type="TEXT">
+                <constraints nullable="true"/>
+            </column>
+            <column name="jdk_vendor" type="TEXT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="commit_sha_short" type="TEXT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="deployed_at" type="DATETIME(6)">
```
mipo256: Cannot we just go for `DATETIME`? I do not think microsends precision is needed here

```suggestion
            <column name="deployed_at" type="DATETIME">
```

master/src/main/resources/db/changelog/db.changelog-master.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+        xsi:schemaLocation="
+            http://www.liquibase.org/xml/ns/dbchangelog
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+    <include file="db/changelog/sqlite/db.changelog-sqlite.xml"/>
```
mipo256: I understand how this is supposed to work. I generally agree with that approach. However, I have two things to say.

1. In the current approach, we would have to add new changeset migrations to the existing changelog XML files. This is going to cause these files to become large in the future, and they are going to become difficult to manage. I suggest having the `db.changelog-postgresql.xml`, which in turn includes the sub-changelogs, where each sub-changelog contains the changes for the release.

2. Current approach assumes that each changeset inside each changelog would have the `rdbms="sqlite"` or something like that filter. If someone would eventually forget to add this to the migration or would copy-paste wrongly, then we're in trouble. I think the best way to solve this problem is to include this master changelog files for each database conditionally, not each individual changeset conditionally.

master/src/test/java/com/axelixlabs/axelix/master/service/state/PostgreSqlDatabaseInstanceRegistryTest.java
```diff
@@ -0,0 +1,32 @@
...
+import com.axelixlabs.axelix.master.utils.db.PostgreSqlTestContainerExtension;
+
+/**
+ * MySQL integration tests for {@link DatabaseInstanceRegistry}.
+ *
+ * @see DatabaseInstanceRegistry
+ * @see PostgreSqlTestContainerExtension
+ * @author Nikita Kirillov
+ */
+@ExtendWith(PostgreSqlTestContainerExtension.class)
```
mipo256: This extension just bootstraps the test container for this test. Where do we set the `spring.datasource.url` or something for us to work with the bootrapped container? I do not see us doing that.

mipo256 (changes requested):
The overall direciton is right, still there are a lot of issues.

PR #917, opened 2026-04-12
GH-253: Integrated all Playground repositories into a monorepo

### LINKED ISSUE #253: Integrate all the playground repos into our own monrepo
It becomes more and more obvious that managing our playgrounds as standalone project becomes more and more complicated. In particular:

1. Once the monorepo is built and the artifacts have been published (see #252), we need to re-deploy our playground with new version of the axile starter. As of now this needs to be done manually for every repo, which is very tedious and error-prone
2. Apart from that, because we use different repos, we have to duplicate various settings, such as docker oauth tokens, or kube config files or etc. So this is also tedious and error-prone etc.

So, the decision has been made for us to integrate playground repos into our own monorepo to simplify the management of those repos and overall deployment. 

I think we can create a separate gradle module called `playground` (similar to the `sbs` module that we have now) that is just going to host other "sub-modules" which in turn would represent our playground repos.

**DOD:** Spring Petclinic, Feature Tracker and Notification Service are migrated to monorepo.

sergeycherkasovv:


.github/workflows/notification-gradle-sb-3.yaml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      NEXUS_USER: ${{ secrets.NEXUS_USER }}
+      NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+      NEXUS_HOST_PORT: ${{ secrets.NEXUS_HOST_PORT }}
+      STORE_PASS: ${{ secrets.STORE_PASS }}
+
+      DOCKER_CR_OAUTH_TOKEN: ${{ secrets.DOCKER_CR_OAUTH_TOKEN }}
+      CR_URL: ${{ vars.CR_URL }}
+      CR_REPOSITORY: ${{ vars.CR_MASTER_REPOSITORY }}
+      KUBE_CONFIG_BASE64: ${{ secrets.KUBE_CONFIG_BASE64 }}
```
sergeycherkasovv, line 31: At the moment, [Composite Run](https://github.com/actions/runner/blob/main/docs/adrs/0549-composite-run-steps.md#secrets) for GitHub Actions has limitations and cannot store `env` variables, so each `playground` has to keep these variables on its own

.github/workflows/petclinic-gradle-sb-2.yaml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      NEXUS_USER: ${{ secrets.NEXUS_USER }}
+      NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+      NEXUS_HOST_PORT: ${{ secrets.NEXUS_HOST_PORT }}
+      STORE_PASS: ${{ secrets.STORE_PASS }}
+
+      DOCKER_CR_OAUTH_TOKEN: ${{ secrets.DOCKER_CR_OAUTH_TOKEN }}
+      CR_URL: ${{ vars.CR_URL }}
+      CR_REPOSITORY: ${{ vars.CR_MASTER_REPOSITORY }}
+      KUBE_CONFIG_BASE64: ${{ secrets.KUBE_CONFIG_BASE64 }}
```
sergeycherkasovv, line 31: At the moment, Composite Run for GitHub Actions has limitations and cannot store env variables, so each playground has to keep these variables on its own

.github/workflows/petclinic-maven-sb-2.yaml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      NEXUS_USER: ${{ secrets.NEXUS_USER }}
+      NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+      NEXUS_HOST_PORT: ${{ secrets.NEXUS_HOST_PORT }}
+      STORE_PASS: ${{ secrets.STORE_PASS }}
+
+      DOCKER_CR_OAUTH_TOKEN: ${{ secrets.DOCKER_CR_OAUTH_TOKEN }}
+      CR_URL: ${{ vars.CR_URL }}
+      CR_REPOSITORY: ${{ vars.CR_MASTER_REPOSITORY }}
+      KUBE_CONFIG_BASE64: ${{ secrets.KUBE_CONFIG_BASE64 }}
```
sergeycherkasovv, line 31: At the moment, Composite Run for GitHub Actions has limitations and cannot store env variables, so each playground has to keep these variables on its own

.github/workflows/feature-maven-sb-3.yaml
```diff
@@ -0,0 +1,57 @@
...
+    env:
+      NEXUS_USER: ${{ secrets.NEXUS_USER }}
+      NEXUS_PASSWORD: ${{ secrets.NEXUS_PASSWORD }}
+      NEXUS_HOST_PORT: ${{ secrets.NEXUS_HOST_PORT }}
+      STORE_PASS: ${{ secrets.STORE_PASS }}
+
+      DOCKER_CR_OAUTH_TOKEN: ${{ secrets.DOCKER_CR_OAUTH_TOKEN }}
+      CR_URL: ${{ vars.CR_URL }}
+      CR_REPOSITORY: ${{ vars.CR_MASTER_REPOSITORY }}
+      KUBE_CONFIG_BASE64: ${{ secrets.KUBE_CONFIG_BASE64 }}
```
sergeycherkasovv, line 31: At the moment, Composite Run for GitHub Actions has limitations and cannot store env variables, so each playground has to keep these variables on its own

mipo256:
N/A, we have already integrated all playgrounds to monorepo

PR #923, opened 2026-04-14
GH-844: Allow self-registration and auto discovery modes to co-exist

### LINKED ISSUE #844: [Feature] Allow self-registration and auto discovery modes to co-exist
This might be needed, for example, in case some services need to be managed by Axelix, but they are deployed from outside this K8S cluster.

**TBD:**

Make sure that Axelix Master can handle both modes at the same time.

We'll also need to introduce the field `latestHearthBeat` in `Instance`. It will never be `null` in the case of self-registered instances, and it will always be `null` in the case of auto-discovered instances. This field must be updated on every received heartbeat.

sergeycherkasovv:


master/src/test/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoverySchedulerTest.java
```diff
@@ -87,7 +87,7 @@ class ShortPollingInstanceDiscoverySchedulerTest {
     @BeforeEach
     void setUp() throws IOException {
         if (mockWebServer != null) {
-            mockWebServer.close();
+            mockWebServer.shutdown();
```
sergeycherkasovv: This change addresses tests failing in **CI** (they pass locally). Tests also pass in **CI debug mode**. I suspect `close()` doesn't properly release resources, causing state to persist between runs and affect subsequent tests.

master/src/main/java/com/axelixlabs/axelix/master/api/internal/endpoint/SelfRegisteredApi.java
```diff
@@ -40,7 +42,11 @@
 @Hidden
 @InternalApiRestController
 @RequestMapping(path = ApiPaths.SelfRegistryApi.MAIN)
-@ConditionalOnProperty(prefix = "axelix.master.discovery", name = "auto", havingValue = "false")
+@ConditionalOnProperty(
+        prefix = "axelix.master.discovery",
+        name = "self-registration",
+        havingValue = "true",
+        matchIfMissing = true)
```
sergeycherkasovv, line 49: Self-registration is enabled by default)

master/src/main/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoveryScheduler.java
```diff
@@ -78,9 +80,15 @@ public void performDiscovery() {
...
+        List<Instance> currentlyRegisteredIds = instanceRegistry.getAll();
 
-        instanceRegistry.reload(discoveredInstances);
+        Set<InstanceId> autoDiscoveryInstancesIds = currentlyRegisteredIds.stream()
+                .filter(instance -> instance.latestHearthBeat() == null)
+                .map(Instance::id)
+                .collect(Collectors.toSet());
+
+        instanceRegistry.deRegisterAll(autoDiscoveryInstancesIds);
+        instanceRegistry.registerAll(discoveredInstances);
```
sergeycherkasovv: I removed the `reload()` method here since we need to delete `Instances` coming from **k8s**. The only concern is `insertAll(...)`, which assumes `InstanceId` uniqueness. In theory, we shouldn't have duplicate `InstanceIds` in the **DB**, but if they appear, we'll get a **DB-level** exception. 
Maybe I'm overthinking it ;)

master/src/main/java/com/axelixlabs/axelix/master/api/internal/endpoint/SelfRegisteredApi.java
mipo256: I think yes, that is generally a good idea.

master/src/test/java/com/axelixlabs/axelix/master/api/internal/SelfRegisteredApiTest.java
```diff
@@ -47,7 +47,7 @@
  * @author Nikita Kirillov
  */
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
-@TestPropertySource(properties = {"axelix.master.discovery.auto=false"})
+@TestPropertySource(properties = {"axelix.master.discovery.self-registration=true"})
```
mipo256:
```suggestion
@TestPropertySource(properties = {"axelix.master.discovery.self-registration.enabled=true"})
```

master/src/test/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoverySchedulerTest.java
mipo256: >  I suspect close() doesn't properly release resources, causing state to persist between runs and affect subsequent tests.

That can be the case I guess, however there are also a couple of things to consider:

[The `close` method actually delegates to `shutdown`](https://github.com/square/okhttp/blob/parent-4.12.0/mockwebserver/src/main/kotlin/okhttp3/mockwebserver/MockWebServer.kt#L913), so there seem to be no reason to assume that `shutdown` will fix the problem.

Apart from that, closing the `MockWebServer` is not related to the Hikari connection pool and the management of the connections inside of it. 

So, I really suspect the problem is in something else. I'll take the proposed change, but I still think that we need to take a deeper look inside this issue. [Created a separate ticket for that](https://github.com/axelixlabs/axelix/issues/944),

master/src/main/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoveryScheduler.java
mipo256: The overall solution I think is alright, but it has two issues:

1. I would not query all the instances that are out there in the database and then filter them in memory like it is currently done. Although the dataset is probably not that large, I would still filter on the database layer to avoid unnecessarily polluting the heap and performance degradation.

2. This entire dance (get/delete/re-insert) with the database must be wrapped inside the single transaction so that we do not suddenly stop in-between the reloading phase, and as a consequence, drop the entire database while not populating it

I would still recommend just containing this complexity inside the `reload`  method. I think just rewriting the reload method and chang its contract on the interface level a bit would be the most viable and concise solution.

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseInstanceRegistry.java
```diff
@@ -64,6 +64,10 @@ public void register(Instance instance) {
         }
     }
 
+    public void registerAll(Collection<Instance> instances) {
```
mipo256: If we consider my comment about the realod re-write, then we would not need it

master/src/main/java/com/axelixlabs/axelix/master/service/state/InstanceRegistry.java
```diff
@@ -47,6 +47,13 @@ public interface InstanceRegistry {
      */
     void register(Instance instance);
 
+    /**
```
mipo256, line 51: If we consider my comment about the realod re-write, then we would not need it

master/src/main/java/com/axelixlabs/axelix/master/api/internal/endpoint/SelfRegisteredApi.java
```diff
@@ -40,7 +42,11 @@
 @Hidden
 @InternalApiRestController
 @RequestMapping(path = ApiPaths.SelfRegistryApi.MAIN)
-@ConditionalOnProperty(prefix = "axelix.master.discovery", name = "auto", havingValue = "false")
+@ConditionalOnProperty(
+        prefix = "axelix.master.discovery",
+        name = "self-registration",
```
mipo256:
```suggestion
        prefix = "axelix.master.discovery.self-registration",
        name = "enabled",
```

mipo256 (changes requested):
So, we're generally headed in the right direction. Still, a couple of issues left.

Please, **pull the changes** (I pushed one small commit), and rebase onto the master branch, since there are some changes that are important there that are for the proposed patch

master/src/test/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoverySchedulerTest.java
sergeycherkasovv: Since you created [issue #944](https://github.com/axelixlabs/axelix/issues/944) to address this problem, I've temporarily disabled the `ShortPollingInstanceDiscoverySchedulerTest` class to keep the build passing)

master/src/main/java/com/axelixlabs/axelix/master/service/discovery/ShortPollingInstanceDiscoveryScheduler.java
sergeycherkasovv: I believe I've addressed your feedback. Please review)

master/src/test/java/com/axelixlabs/axelix/master/api/internal/SelfRegisteredApiTest.java
sergeycherkasovv: Fixed it)

master/src/main/java/com/axelixlabs/axelix/master/service/state/InstanceRegistry.java
sergeycherkasovv: Per your earlier comment, this method is no longer needed after the approach change, so I removed it)

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseInstanceRegistry.java
sergeycherkasovv: Per your earlier comment, this method is no longer needed after the approach change, so I removed it)

master/src/main/java/com/axelixlabs/axelix/master/api/internal/endpoint/SelfRegisteredApi.java
sergeycherkasovv: done)

PR #924, opened 2026-04-15
GH-899 Implement Env & ConfigProps visibility policy

### LINKED ISSUE #899: [Feature] Env & ConfigProps values visibility policy
So, I think I have a pretty solid understanding of how the access control for values in the environment and in `@ConfigurationProperties` feeds will work.

Here are my thoughts about it.

By default, every logged-in user (including `VIEWER`, see #848) is allowed to see the _**"non-sensative"**_ values. Now, imagine company A just adopted Axelix. And here we have ultimately two choices:

1. We can sanitize all the values by default. This means that in order to change the default behavior, **we force** the end users to define their sensitive values if they want to see any values in the environment.

2. Or, we sanitize no values by default, and the end-users see the entire environment by default.

The ultimate problem with the second approach is not just the security issue, the properties can be added to the sanitized list after all. The problem is that by doing the second option, we elongate the process of deploying Axelix which is absolutely not what we want. 

The policy should be simple: **_Safely_ deploy now, customize everything later**.

Now, about RBAC. The sensitive values should be seen by the user who has the appropriate privileges (again, see the RBAC issue). Assuming that the master will proxy the end user's access token to the manager service and Axelix Starter will be able to decode it, I guess it is quite simple: Axelix Starter just decides upon the received token whether to santize the value or not.

NikitaKirilloff:
close #899


mipo256 (approved):
LGTM finally, I've made some changes, but the overall aprroach is fine. Still, we have to think about introducing the shared module for common tests code.

PR #930, opened 2026-04-16
GH-929 Fixed `/axelix-metadata` endpoint error

### LINKED ISSUE #929: [BUG] `/axelix-metadata` returns error when CycloneDX is missing

sergeycherkasovv:


mipo256 (changes requested):
I understand the problem at hand, but I am not really in favor of this approach TBH.

First of all, we have dedicated classes in spring and spring boot like `SpringVersion` and `SpringBootVersion` that can return the versions of spring and spring boot

The version of spring cloud is a complex one since in old spring cloud times like in spring cloud 2020 or smehting [there were no single spring cloud version](https://mvnrepository.com/artifact/org.springframework.cloud/spring-cloud-dependencies/2020.0.4/dependencies). I mean, there were a version of the BOM, but the BOM contained different versions for different components, check yourself. 

That is _maybe_ going to change in the future, at least I hope so. [I am in the process of conversation with Spring Cloud team about it](https://github.com/spring-cloud/spring-cloud-release/issues/451). So for now we just simply cannot say stuff like

> The spring cloud version is 'X'

We can I guess say it if we would somehow magically know the version of the Spring Cloud BOM, but I guess we cannot do it for now.

**Solution:**

So, what I'm actually suggesting is to

1. Get rid of CycloneDX entirely, we do not need it for now.
2. Discover the versions of Spring Framework and Spring Boot via dedicated classes
3. Set Spring Cloud version as empty in all cases. I guess this would be the temporary solution. Add a TODO for this also please

sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultLibraryInformationProvider.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        try {
+            Class<?> kotlinVersionClass = Class.forName("kotlin.KotlinVersion", false, CLASS_LOADER);
+            Object current = kotlinVersionClass.getField("CURRENT").get(null);
+            return String.valueOf(current);
+        } catch (ReflectiveOperationException e) {
+            return null;
+        }
+    }
```
sergeycherkasovv, line 52: Unfortunately, we cannot get the **Kotlin** version using `System.getProperty()` like we do with `java.version`. So I chose this approach)

Initially, I was directly accessing `KotlinVersion` in the `LibraryInformationProvider` interface:
```
default String getKotlinVersion() { 
    try { 
        return KotlinVersion.CURRENT.toString(); 
    } catch (NoClassDefFoundError e) { 
       return null; 
   } 
}
```

Since our library includes **Kotlin**, it may be present on the application's `classpath` when users depend on it. In such cases, we expose the **Kotlin** runtime version detected on the `classpath`—that is, the version of **Kotlin** that may be present in the user's environment. However, this does not necessarily mean that the application itself uses **Kotlin**

However, this approach seemed unreliable to me, so I went with the current solution)

sbs/axelix-spring-boot-3/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultLibraryInformationProvider.java
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        try {
+            Class<?> kotlinVersionClass = Class.forName("kotlin.KotlinVersion", false, CLASS_LOADER);
+            Object current = kotlinVersionClass.getField("CURRENT").get(null);
+            return String.valueOf(current);
+        } catch (ReflectiveOperationException e) {
+            return null;
+        }
+    }
```
sergeycherkasovv, line 52: I've already explained my thoughts above about why I chose this approach)

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/details/InstanceDetailsEndpointTest.java
```diff
@@ -109,50 +106,6 @@ void shouldReturnValidDetailsStructure() {
         assertThatJson(responseBody).inPath("os").isObject().containsOnlyKeys("name", "version", "arch");
     }
 
-    @Test
-    void shouldContainValidDetails() {
```
sergeycherkasovv, line 113: I think this test seems redundant since we already have `shouldReturnValidDetailsStructure`)

sbs/axelix-spring-boot-3/src/test/java/com/axelixlabs/axelix/sbs/spring/core/details/InstanceDetailsEndpointTest.java
```diff
@@ -109,50 +106,6 @@ void shouldReturnValidDetailsStructure() {
         assertThatJson(responseBody).inPath("os").isObject().containsOnlyKeys("name", "version", "arch");
     }
 
-    @Test
-    void shouldContainValidDetails() {
```
sergeycherkasovv, line 113: I think this test seems redundant since we already have `shouldReturnValidDetailsStructure`)

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/AxelixMetadataEndpointTest.java
```diff
@@ -77,6 +75,20 @@ HealthDetectionFunction healthDetectionFunction() {
...
             return () -> "1.1.3";
         }
+
+        @Bean
+        public LibraryInformationProvider libraryInformationProvider() {
+            return new DefaultLibraryInformationProvider();
+        }
+
+        @Bean
+        public ServiceDetailsAssembler serviceDetailsAssembler(
```
mipo256: I do not think we need to declare this as a bean.

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultBasicDiscoveryMetadataAssemblerTest.java
```diff
@@ -73,6 +72,20 @@ HealthDetectionFunction healthDetectionFunction() {
...
             return () -> "1.1.3";
         }
+
+        @Bean
+        public LibraryInformationProvider libraryInformationProvider() {
+            return new DefaultLibraryInformationProvider();
+        }
+
+        @Bean
+        public ServiceDetailsAssembler serviceDetailsAssembler(
```
mipo256: I do not think we need to declare this as a bean.

sbs/axelix-spring-boot-2/src/test/java/com/axelixlabs/axelix/sbs/spring/core/details/DefaultServiceDetailsAssemblerTest.java
```diff
@@ -73,16 +75,16 @@ void shouldAssembleCompleteServiceDetails() {
 
         SpringDetails spring = result.getSpring();
         assertThat(spring).isNotNull();
-        assertThat(spring.getSpringBootVersion()).isNotBlank();
-        assertThat(spring.getSpringFrameworkVersion()).isNotBlank();
+        assertThat(spring.getSpringBootVersion()).as("", SpringBootVersion.getVersion());
```
mipo256: That does not perform any checks. It just provides a description for the assertion, so these `as` are actually a bit meaningless

sbs/axelix-spring-boot-3/src/test/java/com/axelixlabs/axelix/sbs/spring/core/details/DefaultServiceDetailsAssemblerTest.java
```diff
@@ -73,16 +76,16 @@ void shouldAssembleCompleteServiceDetails() {
 
         SpringDetails spring = result.getSpring();
         assertThat(spring).isNotNull();
-        assertThat(spring.getSpringBootVersion()).isNotBlank();
-        assertThat(spring.getSpringFrameworkVersion()).isNotBlank();
+        assertThat(spring.getSpringBootVersion()).as("", SpringBootVersion.getVersion());
```
mipo256: Same stuff here - these `as` calls just provide the description. They do not check anything.

sbs/axelix-spring-boot-3/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/AxelixMetadataEndpointTest.java
```diff
@@ -74,9 +72,26 @@ HealthDetectionFunction healthDetectionFunction() {
         }
 
         @Bean
-        AxelixVersionDiscoverer axelixVersionDiscoverer() {
+        AxelixVersionDiscoverer axelixVersionDiscoverer(
+                GitInformationProvider gitInformationProvider,
```
mipo256: This `AxelixVersionDiscoverer` does not need any of those method arguments.

sbs/axelix-spring-boot-3/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultServiceMetadataAssemblerTest.java
```diff
@@ -78,6 +77,20 @@ HealthDetectionFunction healthDetectionFunction() {
...
             return () -> "1.1.3";
         }
+
+        @Bean
+        public LibraryInformationProvider libraryInformationProvider() {
+            return new DefaultLibraryInformationProvider();
+        }
+
+        @Bean
+        public ServiceDetailsAssembler serviceDetailsAssembler(
```
mipo256: We do not seem to need this bean here.

sbs/axelix-spring-boot-3/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/AxelixMetadataEndpointTest.java
```diff
@@ -74,9 +72,26 @@ HealthDetectionFunction healthDetectionFunction() {
...
             return () -> "1.1.3";
         }
+
+        @Bean
+        public LibraryInformationProvider libraryInformationProvider() {
+            return new DefaultLibraryInformationProvider();
+        }
+
+        @Bean
+        public ServiceDetailsAssembler serviceDetailsAssembler(
```
mipo256: We do not seem to need this bean at all

sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultLibraryInformationProvider.java
mipo256: I just realized that we have a greater problem than that. We actually bring Kotlin standard library inside the user's classpath. 

This is generally bad since we might conflict with the version that the end-user acutally wants to bring into the classpath. But this is another problem that does not pertain to your issue. 

So, yeah, we're just bringing the entire Kotlin standard library into the end-users classpath for basically a very small amount of classes. I am not sure that this is a good idea. But in any case, let's leave it aside for now.

mipo256 (changes requested):
The changes are generally good, there are a couple of minor issue.

The bigger concern is of course the fact that we bring kotlin to the classpath of users and whether this is a good thing. I think it is fine... For now.

PR #945, opened 2026-04-20
GH-918 Handle the role decoding capabilities from OIDC tokens

### LINKED ISSUE #918: [Feature] Handle the role decoding capabilities from OIDC tokens
Currently, every user that has logged in via OIDC has the `ADMIN` role which was a temporary measure. 

Before release, we have to provide a way to end-users to specify from where, actually, the role should be taken. Probably, it should be taken from the specific claim of the JWT access token.

NikitaKirilloff:
close #918 

## Summary by CodeRabbit

* **New Features**
  * Dynamic role assignment from OAuth2 tokens using configurable JMESPath expressions; falls back to user-info when needed.
  * OAuth2 flow now handles both ID and access tokens to improve identity and role resolution.
  * Extracted role strings map to admin/editor/viewer with a safe default of viewer.

* **Configuration Changes**
  * OAuth2 property `redirectUri` replaced by `baseUrl` (redirect is now derived).
  * Added optional `roleAttributePath` to configure role extraction.

### DIFF
--- a/master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
@@ -17,9 +17,6 @@
  */
 package com.axelixlabs.axelix.master.autoconfiguration.auth.properties;
 
-import java.util.List;
-import java.util.Map;
-
 import org.jspecify.annotations.Nullable;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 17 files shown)


master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/OAuth2Properties.java
```diff
@@ -44,22 +50,33 @@ public record OAuth2Properties(
         String issuerUri,
         String clientId,
         String clientSecret,
-        String redirectUri,
+        String baseUrl,
         @Nullable String usernameClaim,
-        String scopes) {
+        String scopes,
+        @Nullable String roleAttributePath,
+        Map<String, List<String>> roleMapping) {
```
NikitaKirilloff: If roleAttributePath is not specified in the application.property, always fallback to DefaultRole.VIEWER. 

See class OIdcRoleExtractor.

master/src/test/java/com/axelixlabs/axelix/master/service/auth/oauth/OidcRoleExtractorTest.java
NikitaKirilloff, line 1: hmm... I'm not sure we need that many tests, because essentially I'm testing a library `com.jayway.jsonpath:json-path`  here. WDYT @mipo256 ?

mipo256: I actually think that is fine, since it is an abstract after all that has some API. So the API must be tested, since the implementation behind the abstraction might be different.

mipo256:
I've decided to use the jmes-path library, which is archived now. This is generally, of course, not a good idea, but still I think it is better than JsonPath for various reasons discussed offline. In any case, if needed, we will be able to just embed this code into our monorepo like spring does with `cglib`, and just support it ourselves. 

CC: @NikitaKirilloff @sergeycherkasovv

PR #962, opened 2026-04-23
GH-938: added Users Management API logic

### LINKED ISSUE #938: [Feature] Design the backend API for creating the users
This API is supposed (at least as we see now) to be used by the JavaScrpipt front-end when creating users from the UI. It is also supposed to be used by the Terraform provider to create users via as-a-code approach.

We also need to design the table in the database for storing users. 

**Spec (on top of my head):**

1. Accepts the username, password, and the role to be assigned. 
2. Returns just 200 OK status code

As a result, the user is created inside the dedicated table with the password hashed prior to that (i.e. prior to storing it in the DB).

**IAM:**

Only `SUPER_ADMIN` is allowed to create users, i.e. the `SUPER_ADMIN` roel would be the only role in the Axelix Community Edition that would contain the permission to do this.

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Secure password encryption for user logins using BCrypt
  * Last login timestamp tracking for users
  * Functional user management endpoints (create, delete, update users)
  * Support for partial user updates with optional fields
  * Optional password field during user creation

* **Bug Fixes**
  * User login now validates credentials against encrypted passwords


master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/PersistenceAutoConfiguration.java
```diff
@@ -213,7 +217,46 @@ public static class BaseJdbcConvertersAutoConfiguration extends AbstractJdbcConf
...
+            }
+
+            @Override
+            public UserEntity.Roles convert(String source) {
+                if (source == null || source.isBlank()) {
+                    return new UserEntity.Roles(Set.of());
+                }
+                Set<String> roles = jsonMapper.readValue(source, new TypeReference<>() {});
+                return new UserEntity.Roles(roles);
+            }
```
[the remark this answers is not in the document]
sergeycherkasovv: This issue can be ignored for now)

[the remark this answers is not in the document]
sergeycherkasovv: @coderabbitai resolve

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/RoleNotPresentException.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+public class RoleNotPresentException extends RuntimeException {
+
+    public RoleNotPresentException(@Nullable String role) {
+        super("Role '%s' is not allowed".formatted(role));
+    }
+
+    public RoleNotPresentException(final String role, final Throwable cause) {
+        super(role, cause);
+    }
```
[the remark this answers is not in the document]
sergeycherkasovv: @coderabbitai resolve

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserIdNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class UserIdNotFoundException extends RuntimeException {
+
+    public UserIdNotFoundException(String id) {
+        super("User with id '%s' is not found".formatted(id));
+    }
+
+    public UserIdNotFoundException(final String id, final Throwable cause) {
+        super(id, cause);
+    }
```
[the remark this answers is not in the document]
sergeycherkasovv: @coderabbitai resolve

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
```diff
@@ -0,0 +1,140 @@
...
+    @Override
+    public Optional<UserEntity> getUserById(String id) {
+        return userRepository.findById(id);
+    }
+
+    @Override
+    public void updateLastLoginAt(String username) {
+        userRepository
+                .findByUsername(username)
+                .ifPresent(user -> jdbcAggregateTemplate.update(user.withLastLoginAt(Instant.now())));
```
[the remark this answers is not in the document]
sergeycherkasovv: This issue requires discussion with the team, as we haven't yet discussed the chosen approach)

master/src/main/java/com/axelixlabs/axelix/master/service/export/StateArchiveFileNameGenerator.java
```diff
@@ -21,7 +21,7 @@
 import com.axelixlabs.axelix.master.domain.InstanceId;
 
 /**
- * Abstraction that is capable to generate the file name for the state file archive.
+ * Abstraction that is capable to generate the file username for the state file archive.
```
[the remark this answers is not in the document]
sergeycherkasovv: @coderabbitai resolve

master/src/main/java/com/axelixlabs/axelix/master/api/external/request/user/UserCreateRequest.java
```diff
@@ -22,17 +22,19 @@
...
+ * @param username   Login name of the user to create.  Must be unique.
+ * @param email      The user email address, which may be {@code null}. Must be unique.
+ * @param password   Plain-text password , which may be {@code null}. Hashed server-side before persistence.
+ * @param role       Name of the role to grant to the user.
  *
  * @author Sergey Cherkasov
  */
-public record UserCreateRequest(String name, @Nullable String email, String password, String role) {
+public record UserCreateRequest(
+        String username, @Nullable String email, @Nullable String password, String role) {
```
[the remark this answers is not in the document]
sergeycherkasovv: We need to discuss this policy within the team first :)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
```diff
@@ -96,19 +107,24 @@ public ResponseEntity<?> login(@RequestBody LoginRequest loginRequest) {
...
+        String token;
 
-            return ResponseEntity.status(HttpStatus.OK)
-                    .header(HttpHeaders.SET_COOKIE, cookie.toString())
-                    .header(HttpHeaders.SET_COOKIE, cookieAuthorities.toString())
-                    .build();
+        if (Objects.equals(user.getPassword(), loginRequest.password())) {
+            token = jwtEncoderService.generateToken(user);
+        } else if (passwordEncoder.matches(loginRequest.password(), user.getPassword())) {
+            token = jwtEncoderService.generateToken(user);
```
NikitaKirilloff: To be honest, I don’t understand why we check the password twice here.

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserManagementApi.java
```diff
@@ -79,23 +85,56 @@ public ResponseEntity<List<UserResponse>> getUsersFeed() {
...
 
     @DefaultApiResponse(summary = "Update a user")
     @ApiResponse(description = "No Content", responseCode = "204")
     @PatchMapping(path = ApiPaths.UsersManagementApi.USERS_UPDATE)
     public ResponseEntity<Void> updateUser(@RequestBody UserUpdateRequest request) {
+        try {
+            if (request.roles() != null) {
+                if (request.roles().stream().anyMatch(role -> role == null || role.isBlank())) {
+                    return ResponseEntity.status(HttpStatus.BAD_REQUEST).build();
+                }
```
NikitaKirilloff: I think this can be moved to a separate validation private method.

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
NikitaKirilloff: It's very strange that we loaded the user and then validated the password below. I might be wrong, but we should validate the password directly in the StaticAdminUserProvider.
or some UserService

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
```diff
@@ -96,19 +107,24 @@ public ResponseEntity<?> login(@RequestBody LoginRequest loginRequest) {
...
-            ResponseCookie cookie = cookieService.buildAuthCookie(token);
-            ResponseCookie cookieAuthorities = cookieService.buildAuthoritiesMetadataCookie(user.getRoles());
+        String token;
 
-            return ResponseEntity.status(HttpStatus.OK)
-                    .header(HttpHeaders.SET_COOKIE, cookie.toString())
-                    .header(HttpHeaders.SET_COOKIE, cookieAuthorities.toString())
-                    .build();
+        if (Objects.equals(user.getPassword(), loginRequest.password())) {
+            token = jwtEncoderService.generateToken(user);
```
NikitaKirilloff: I think the logic for checking the correctness of the password shouldn't be in the controller, there should be some UserService

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/StaticAdminUserProvider.java
```diff
@@ -46,8 +52,26 @@ public User load(String username) throws UserNotFoundException {
...
+        UserEntity user =
+                userManaged.getUserByUsername(username).orElseThrow(() -> new UserNotFoundException(username));
+
+        if (user.password() == null) {
             throw new UserNotFoundException(username);
         }
+
+        return new DefaultUser(
+                user.username(), user.password(), extractRoles(user.roles().values()));
+    }
```
NikitaKirilloff: This class is doing something weird right now. Judging by its name, StaticAdminUserProvider, it should only work with the StaticAdmin entity, i.e., with a user with the SUPER_ADMIN role. It shouldn't work through some service with a hypothetical UserRepository.

master/src/main/java/com/axelixlabs/axelix/master/service/state/UserManaged.java
```diff
@@ -0,0 +1,114 @@
...
+ * <p>Provides operations to create, read, update, and delete {@link UserEntity} records,
+ * as well as to track login activity. Passwords supplied to this service are plain-text and
+ * are hashed by the implementation before persistence. Usernames and emails are expected to
+ * be unique; attempts to violate those constraints surface as
+ * {@link org.springframework.dao.DataIntegrityViolationException}.
+ *
+ * @author Sergey Cherkasov
+ */
+public interface UserManaged {
+
```
NikitaKirilloff: The name of this class doesn't quite make it clear what it does or why it exists. And just looking at its name, it's unclear where it fits in the proposed Authentication Request Hierarchy.
`@Controller` -> `@Service` -> `@Repository`
I'm not sure if this is correct, but maybe we should call it UserService ? WDYT @sergeycherkasovv 

I also understand that the implementation of this interface, i.e. the service `@Service` will work with JdbcTemplate, but I think we can allow this

NikitaKirilloff (changes requested):
I think we need to rework the structure. I'm not sure we need to put the StaticAdminUserProvider class in a UserService, since this UserService should only exist when users have login/password authentication enabled. I think we need to rethink the entire structure.

Let's discuss this. @sergeycherkasovv 

Perhaps we need an implementation of the strategy pattern e.g. `List<UserProvider> userProviders;` I'm not sure, but think about it

e.g.
``` 
OidcUserProvider implement UserProvider
StaticAdminUserProvider implement UserProvider
BasicAuthUserProvider implement UserProvider
```
We need to discuss this with the team.

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserIdNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.axelixlabs.axelix.master.repository.UserRepository;
+
+/**
+ * Thrown in case the user id is not found by {@link UserRepository}.
+ *
+ * @see UserRepository
+ * @author Sergey Cherkasov
+ */
+public class UserIdNotFoundException extends RuntimeException {
```
NikitaKirilloff:
```suggestion
public class UserNotFoundException extends RuntimeException {
```

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserManagementApi.java
```diff
@@ -79,23 +85,56 @@ public ResponseEntity<List<UserResponse>> getUsersFeed() {
     @ApiResponse(description = "Created", responseCode = "201")
     @PostMapping(path = ApiPaths.UsersManagementApi.USERS_CREATE)
     public ResponseEntity<Void> createUser(@RequestBody UserCreateRequest request) {
+        try {
+            String role = request.role();
+            if (role == null || role.isBlank() || role.trim().toUpperCase().equals(DefaultRole.SUPER_ADMIN.getName())) {
```
NikitaKirilloff: implementation("org.springframework.boot:spring-boot-starter-validation") WDYT @mipo256

sergeycherkasovv: During the call, it was decided not to add this dependency to the project for now. We may add it in the future, but for now we handle validation ourselves)

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserIdNotFoundException.java
sergeycherkasovv: I think it would be better to add a separate exception. I chose `UserWithIdNotFoundException`, the name you suggested during our call)

master/src/main/java/com/axelixlabs/axelix/master/service/state/UserManaged.java
sergeycherkasovv: Thank you for the suggestion, I agree with you and renamed it to `UserService`

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/StaticAdminUserProvider.java
sergeycherkasovv: Yes, you're right, and during the call it was decided to follow the **Strategy** pattern you proposed—that is, to separate responsibilities: `StaticAdminUserProvider` only handles the `SUPER_ADMIN` specified in the configuration, while `DatabaseUserProvide`r only handles `Users` that exist only within the **DB**. These services are injected into one **Controller**)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
sergeycherkasovv: In our case, the `DatabaseUserProvider` service is responsible for this logic)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
sergeycherkasovv: Yes, you're right, this logic was moved to `StaticAdminUserProvider` and `DatabaseUserProvider`)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserManagementApi.java
sergeycherkasovv: This logic was moved to `UserService`)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApi.java
sergeycherkasovv: We got rid of it :)

master/src/main/java/com/axelixlabs/axelix/master/api/external/request/user/UserCreateRequest.java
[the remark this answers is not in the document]
sergeycherkasovv: We assume that `SUPER_ADMIN` doesn't want to delete a `User` but still wants to guarantee preventing the user from accessing the system, so we allow `SUPER_ADMIN` to set the password to null for a specific `User`, even if that User was created in our system (`LOCAL`)

common/auth/src/main/java/com/axelixlabs/axelix/common/auth/core/DefaultRole.java
```diff
@@ -34,6 +34,7 @@
  */
 public final class DefaultRole implements Role {
 
+    public static final Set<String> ALL_ROLE_NAMES;
```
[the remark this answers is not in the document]
NikitaKirilloff: I think coderabbit right, the field name is a bit misleading

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/DatabaseUserProvider.java
```diff
@@ -0,0 +1,73 @@
...
+        this.passwordEncoder = passwordEncoder;
+    }
+
+    @Override
+    public @Nullable User load(String username, String password) {
+
+        UserEntity user = userService.getUserByUsername(username).orElse(null);
+
+        if (user != null && user.password() != null && passwordEncoder.matches(password, user.password())) {
+            userService.updateLastLoginAt(user.username());
```
NikitaKirilloff, line 58: Perhaps the userService.updateLastLoginAt(user.username()); logic should be done asynchronously, and perhaps errors during updates should be ignored. But I'm not sure, I'm curious what @mipo256  thinks about this.

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
```diff
@@ -0,0 +1,152 @@
...
+/**
+ * JDBC-based implementation of {@link UserService} that persists users in a relational database.
+ *
+ * @author Sergey Cherkasov
+ */
+@Service
+@NullMarked
+@Transactional
+public class DatabaseUserService implements UserService {
+
```
NikitaKirilloff, line 51: This doesn't apply to the current PR, I think it's worth revising the name of the DatabaseUserService class to DefaultUserService. It's unlikely that we'll have other services for storing users. similarly DatabaseInstanceRegistry

master/src/main/java/com/axelixlabs/axelix/master/service/auth/Provider.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.axelixlabs.axelix.master.service.auth;
+
+/**
+ * Origin of a managed user account.
+ *
+ * @author Sergey Cherkasov
+ */
+public enum Provider {
+    OIDC("OIDC/OAUTH2"),
```
NikitaKirilloff: maybe it's worth changing the name to LoginProvider, or something similar, I'm not sure.

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/UserProvider.java
```diff
@@ -17,23 +17,26 @@
  */
 package com.axelixlabs.axelix.master.service.auth.provider;
 
+import org.jspecify.annotations.Nullable;
+
 import com.axelixlabs.axelix.common.auth.core.User;
-import com.axelixlabs.axelix.master.exception.auth.UserNotFoundException;
 
 /**
  * SPI interface that is capable to load the {@link User} by his/her username.
```
mipo256: Javadoc is not correct here anymore.

master/src/main/java/com/axelixlabs/axelix/master/service/auth/Provider.java
mipo256: I'd just call it `UserOrigin`, since this is de-factor the origin of the given user.

But yeah, the name _"Provider"_ is certainly going to be misleading, accounting for the fact that we have `UserProvider` in the system, which is the abstraction that is not really closely related to this `enum`.

I'd also add the javadoc on top of the `LOCAL` enum entry specifying what do we mean by `LOCAL`, since this is not obvious.

master/src/main/java/com/axelixlabs/axelix/master/service/auth/Provider.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final String display;
+
+    Provider(String display) {
+        this.display = display;
+    }
+
+    /**
+     * Returns the human-readable label of this provider, suitable for display in the UI.
+     */
```
mipo256: 1. Javadoc should be over the field and not the getter accessor, because the `displayName` field can be used in multiple functions, and it would be clear what the `getDisplayName` getter returns, it is just the accessor. Still, it is more interesting what `displayNam ` actually is.

2. "of this provider" --> "of this origin"
3. `display` --> `displayName`

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
mipo256: I think this is relatively inconsequential TBH. Let's leave it as it is. The name `DatabaseUserService` provides some hint about its implementation, which may be good for the reader. Again, this is less of a concern, let's leave it as it is.

common/auth/src/main/java/com/axelixlabs/axelix/common/auth/core/DefaultRole.java
```diff
@@ -65,6 +66,8 @@ public final class DefaultRole implements Role {
                 "SUPER_ADMIN", Arrays.stream(DefaultAuthority.values()).collect(Collectors.toUnmodifiableSet()));
 
         MANAGED_SERVICE = new DefaultRole("MANAGED_SERVICE", Set.of(SELF_REGISTER_AUTHORITY));
+
+        ALL_ROLE_NAMES = Set.of(VIEWER.getName(), ADMIN.getName(), EDITOR.getName());
```
mipo256, line 71: That is a bad solution for the problem.

We have at hand is that we have a set of roles in the system, let's name it _"A"_, and when creating the new user, we want to allow the client to specify the role from some sub-set of _"A"_ - let's call it _"B"_. The idea is good, and this is a valid concern. 

The problem is that if this set of roles grows, you will not be able to manage them via `ALL_ROLE_NAMES`. The developer may simply forget to add the new role into `ALL_ROLE_NAMES`, and we will have the bug, if this is not covered. by tests.

Furthermore, what if there will be some roles in the set _"B"_ in the enterprise edition that are not in the community edition (and this will happen we know it)? The `ALL_ROLE_NAMES` is static and is not configurable. Making it mutable `Set` is actually a security exploit. 

**Solution:**

So what we want is the solution, that, among other things:

1. Maintainable, in the sense that we will need to be able to add roles into this _"B"_ set from outside (e.g. enterprise edition).
2. And you want some 

So, just introduce the `RoleConfig` model or something like that, and then just add the `getRoleConfig` to the `Role` interface. The `RoleConfig` model must include the property like `isManageable`, that will return `true` if we can create users with this role or assign this role to existing users.

The naming can be changed a bit, but I hope the overall point is clear. Now, I invite you to think about why this approach solves two problems above.

master/src/main/java/com/axelixlabs/axelix/master/api/external/request/user/UserCreateRequest.java
```diff
@@ -22,17 +22,19 @@
...
 
     @Override
     public String toString() {
-        return "UserCreateRequest[name=" + name + ", email=" + email + ", password=[REDACTED], role=" + role + ']';
+        return "UserCreateRequest[username=" + username + ", email=" + email + ", password=[REDACTED], role=" + role
+                + ']';
```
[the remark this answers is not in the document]
mipo256: Coderabbit is right @sergeycherkasovv

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserManagementApi.java
```diff
@@ -79,23 +85,44 @@ public ResponseEntity<List<UserResponse>> getUsersFeed() {
     @ApiResponse(description = "Created", responseCode = "201")
     @PostMapping(path = ApiPaths.UsersManagementApi.USERS_CREATE)
     public ResponseEntity<Void> createUser(@RequestBody UserCreateRequest request) {
+        try {
 
-        return ResponseEntity.status(HttpStatus.CREATED).build();
+            userService.create(request.username(), request.email(), request.password(), request.role(), Provider.LOCAL);
+            return ResponseEntity.status(HttpStatus.CREATED).build();
+
+        } catch (UserInvalidValueException
```
mipo256, line 93: I think we should not do stuff like that. We have the `ExceptionHandlingFilter` for that. Still, there are some issues with exception handling on the `Master` side. I will look into that once I'd have some time

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/UserManagementApi.java
```diff
@@ -79,23 +85,44 @@ public ResponseEntity<List<UserResponse>> getUsersFeed() {
...
+    @PutMapping(path = ApiPaths.UsersManagementApi.USERS_UPDATE)
     public ResponseEntity<Void> updateUser(@RequestBody UserUpdateRequest request) {
+        try {
+            userService.updateUserPatch(
+                    request.id(), request.username(), request.email(), request.password(), request.roles());
 
-        return ResponseEntity.noContent().build();
+            return ResponseEntity.noContent().build();
+
+        } catch (UserInvalidValueException
```
mipo256, line 120: Exceptions: ditto as api above

master/src/main/java/com/axelixlabs/axelix/master/api/external/response/UserResponse.java
```diff
@@ -41,11 +42,11 @@ public record UserResponse(
         String username,
         @Nullable String email,
         Set<String> roles,
-        String provider,
+        Provider provider,
```
mipo256: Am I correct that we're essentially returning the part of the persistence model from the API? Like, the `Provider` (to be renamed to `UserOrigin`) is used in the `UserEntity` model, AFAICT. 

If this is the case, then this is the dead end. We should not do that. In DTO it is absolutely okay to return `String`. If we want some additional type safety, we introduce the enum in the `response` package, but I think this is redundant. I'd just go with `String` for now.

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/SecurityAutoConfiguration.java
```diff
@@ -154,6 +164,11 @@ public StaticAdminUserProvider staticCredentialsUserProvider(
                 StaticAdminCredentialsProperties staticCredentialsConfig) {
             return new StaticAdminUserProvider(staticCredentialsConfig);
         }
+
+        @Bean
+        public DatabaseUserProvider databaseUserProvider(UserService userService, PasswordEncoder passwordEncoder) {
```
mipo256, line 169: Let's leave this bean in the `StaticCredentialsConfig` for now, but later we would need to re-think the structure of the configuration. Since `"static-admin"` is not really the auth option anymore and credentials for static admin must not reside in the `options` config.

Anyway, this is for another task. I'll create an issue for that.

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserInvalidValueException.java
```diff
@@ -0,0 +1,35 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.exception.auth;
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * Thrown in case one of the provided user field values is invalid.
```
mipo256: Are we sure that the javadoc is correct here? I guess it is not.

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/UserProvider.java
```diff
@@ -17,23 +17,26 @@
...
      * Load user by username.
      *
      * @param username by which the user will be loaded.
+     * @param password the user's password.
      * @return the loaded {@link User}.
-     * @throws UserNotFoundException in case the user is not found.
      */
-    User load(String username) throws UserNotFoundException;
+    @Nullable
+    User load(String username, String password);
```
mipo256, line 41: I guss I would need remove the `UserNotFoundException` and related stuff.

**IMPORTANT:** This is for future polishing; do not make any actions in this PR

master/src/main/java/com/axelixlabs/axelix/master/api/external/request/user/UserCreateRequest.java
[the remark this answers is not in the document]
mipo256: > We assume that SUPER_ADMIN doesn't want to delete a User but still wants to guarantee preventing the user from accessing the system, so we allow SUPER_ADMIN to set the password to null for a specific User, even if that User was created in our system (LOCAL)

Guys, that is a wrong view of the situation, and we've talked about it. In order to disallow access to the system while preserving the user, we need to introduce the status model for the `UserEntity`. For now, we do not have it, and we do not need it.

The password must always be there for `LOCAL` users, **always**.

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserWithIdNotFoundException.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.axelixlabs.axelix.master.repository.UserRepository;
+
+/**
+ * Thrown in case the user id is not found by {@link UserRepository}.
+ *
+ * @see UserRepository
+ * @author Sergey Cherkasov
+ */
+public class UserWithIdNotFoundException extends RuntimeException {
```
mipo256: Lets not overcomplicate the name
```suggestion
public class UserNotFoundException extends RuntimeException {
```

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/DatabaseUserProvider.java
```diff
@@ -0,0 +1,73 @@
...
+import com.axelixlabs.axelix.master.domain.UserEntity;
+import com.axelixlabs.axelix.master.exception.auth.UserRoleNotFoundException;
+import com.axelixlabs.axelix.master.service.state.UserService;
+
+/**
+ * {@link UserProvider} that authenticates a given user against the users stored in the database.
+ *
+ * @author Sergey Cherkasov
+ */
+public class DatabaseUserProvider implements UserProvider {
```
mipo256, line 42: Where is transaction management, guys?

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/DatabaseUserProvider.java
mipo256: I think this is clearly an overengineering. We do not need to dispatch a new thread for this, there is not a bottleneck for now, and there probably will not be. So let's not overcomplicate the system, we do not even have the problem at hand.

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
```diff
@@ -0,0 +1,152 @@
...
+        return userRepository.findByUsername(username);
+    }
+
+    @Override
+    public Optional<UserEntity> getUserById(String id) {
+        return userRepository.findById(id);
+    }
+
+    @Override
+    public void updateLastLoginAt(String username) {
```
mipo256, line 102: This should be just a single `@Query` with the `UPDATE` statement in it. We're making two queries for no reason here.

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
```diff
@@ -0,0 +1,152 @@
...
+        userRepository
+                .findByUsername(username)
+                .ifPresent(user -> jdbcAggregateTemplate.update(user.withLastLoginAt(Instant.now())));
+    }
+
+    @Override
+    public void updateUserPatch(
```
mipo256, line 107: This should also be just a single `UPDATE` SQL statement.

Now, I know that I've said that writing raw SQL should generally be avoided, and that is still the case for the overall baseline

However, here, if we will replace this appraoch with raw SQL

1. We will have just a single SQL query, instead of **N queries (!)** for fetching the entire aggregate into memory. And hten yet another **N queries** for updating it.
2. The code will look much, mcuh simpler
3. If we apply Srping Data AOT Repositories, we will alleviate some pain of validating queries at compile time, so the invalid SQL queries in their majority will simply not compile. 

So again, in each individual case this must be thought-through, and here it unfortunately was not.

mipo256 (changes requested):
This PR has a large area for improvement. Please, see below

common/auth/src/main/java/com/axelixlabs/axelix/common/auth/core/DefaultRole.java
sergeycherkasovv: It was decided to keep `ALL_ROLE_NAMES` for now as a temporary solution, but a [task](https://github.com/axelixlabs/axelix/issues/971) was also created where we will remove `ALL_ROLE_NAMES`

master/src/main/java/com/axelixlabs/axelix/master/service/auth/Provider.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/UserProvider.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/service/auth/Provider.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/api/external/response/UserResponse.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserInvalidValueException.java
sergeycherkasovv: I fixed it, please take a look :)

master/src/main/java/com/axelixlabs/axelix/master/service/auth/provider/DatabaseUserProvider.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseUserService.java
sergeycherkasovv: I fixed it, but we need to consider that we now have two `UPDATE`  SQL statement:
1. `updateUserPatchWithoutPassword` — used when the password should remain unchanged
2. `updateUserPatch` — used when the password needs to be updated

master/src/main/java/com/axelixlabs/axelix/master/api/external/request/user/UserCreateRequest.java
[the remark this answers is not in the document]
sergeycherkasovv: I made the changes suggested by **CodeRabbit** for `UserResponse`, `UserCreateRequest`, `UserUpdateReques`t, and `UserEntity`)

master/src/main/java/com/axelixlabs/axelix/master/exception/auth/UserWithIdNotFoundException.java
sergeycherkasovv: I got rid of `UserWithIdNotFoundException` entirely  and `UserNotFoundException` is not used in this task)

PR #1010, opened 2026-05-10
GH-1005: Use Spring ProxyFactory to enhance CacheManager beans

### LINKED ISSUE #1005: [Feature] Change CacheManagerBeanPostProcessor implementation. From decorating to proxying
Currently, we're "proxing" the `CacheManager` instances wrong. What we're essentially doing is that [we're decorating over](https://github.com/axelixlabs/axelix/blob/master/sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/cache/CacheManagerBeanPostProcessor.java#L45) the `CacheManager` instance and then delegating the calls to underlying implementation.

Doing so has one major drawback - all caches in the user's classpath will now be instances of `EnhancedCacheManager`, which is not the child of the end user's CacheManager like `CaffeineCacheManager` or `RedisCacheManager`. Thus, checks like `cacheManager instanceOf CaffeineCacheManager` or similar will simply fail.

Furthermore, and this one is probably the most dangerous one, **any kind of injection that is not by `CacheManager` interface will fail**. 

We should probably use the Spring's `ProxyFactory`abstraction to create proxies [as we already do with transactions tracking](https://github.com/axelixlabs/axelix/blob/master/sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/TransactionMonitoringBeanPostProcessor.java#L140).

chrshnv:
Replace the manual EnhancedCacheManager decorator with a CGLIB proxy

Closes #1005 

## Summary by CodeRabbit

* **Refactor**
  * Cache managers are now exposed as enhanced cache contracts via AOP-backed proxies for consistent runtime behavior.
* **New Features**
  * Added a default enhanced cache manager with lazy per-cache wrapping, cache listing, clear/evict, and per-cache/global enable/disable controls.
* **Tests**
  * Unit tests updated to use the new default implementation and adjusted expectations for cache enablement.

[Review Change Stack](https://app.coderabbit.ai/change-stack/axelixlabs/axelix/pull/1010)

### DIFF
--- a/sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/cache/CacheManagerBeanPostProcessor.java
+++ b/sbs/axelix-spring-boot-2/src/main/java/com/axelixlabs/axelix/sbs/spring/core/cache/CacheManagerBeanPostProcessor.java
@@ -37,7 +37,7 @@ public class CacheManagerBeanPostProcessor implements BeanPostProcessor {
 
     @Override
     public Object postProcessAfterInitialization(@NonNull Object bean, @NonNull String beanName) throws BeansException {
-        if (!(bean instanceof CacheManager) || bean instanceof EnhancedCacheManager) {
+        if (!(bean instanceof CacheManager)) {
             return bean;
         }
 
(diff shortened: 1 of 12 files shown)


mipo256:
In general, we try to make PRs less than 500 lines, but that is okay for now, thanks @chrshnv !

Do you mind reviewing/fixing the comments left by code-rabbit ai?

chrshnv:
> In general, we try to make PRs less than 500 lines, but that is okay for now, thanks @chrshnv !
> 
> Do you mind reviewing/fixing the comments left by code-rabbit ai?

yeah, sure

chrshnv:
@coderabbitai review

mipo256 (approved):
Overall pretty good, thanks @chrshnv!

I'll polish it myself, I guess. In general, the only things I say is that:

1. It would be nice to have tests that verify the actual
2. The `enableAll`/`disableAll`/`clearAll` semantics in regards to `getAll` is indeed a restriction of the implementation that we cannot avoid. I will add this to javadoc of the interface. We had this restriction in mind, we just did not explicitly state that, which is a mistake, I guess.

PR #1018, opened 2026-05-11
GH-944 Stabilized the test inside ShortPollingInstanceDiscoverySchedulerTest

### LINKED ISSUE #944: [Feature] Test inside ShortPollingInstanceDiscoverySchedulerTest is flaky
The `ShortPollingInstanceDiscoverySchedulerTest` is flaky. The test method `shouldRegisterNewK8sInstancesWhenDiscovered()` _sometimes_ fails with the exception of having contention for the shared SQLite memory database. This problem sometimes fails the CI build, as well as local builds.

I highly suspect that the problem is that some tests are actually getting executed by Gradle in some specific order (or maybe in parallel inside the test class, idk), and that is the reason why this specific test method in `ShortPollingInstanceDiscoverySchedulerTest` fails, and not the other test methods in other classes.

sergeycherkasovv:
## Summary by CodeRabbit

* **Tests**
  * Re-enabled a previously skipped test for instance discovery functionality.

* **Refactor**
  * Simplified internal logic and reorganized import statements across codebase components.

* **Style**
  * Updated code formatting and indentation for improved readability.

### DIFF
--- a/master/src/test/resources/application.yaml
+++ b/master/src/test/resources/application.yaml
@@ -9,7 +9,7 @@ spring:
     # what we want since we re-use the connections from the Hikari CP inside the single test (setup code relies that the
     # changes it made will be seen by the actual test code). So, we use 'cache=shared'
     # in order to force all the connections to use the shared database.
-    url: "jdbc:sqlite:file:memdb?mode=memory&cache=shared"
+    url: "jdbc:sqlite:file:memdb_${random.uuid}?mode=memory&cache=shared"
   ai:
     mcp:
       server:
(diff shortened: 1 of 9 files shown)


master/src/test/resources/application.yaml
```diff
@@ -9,7 +9,7 @@ spring:
     # what we want since we re-use the connections from the Hikari CP inside the single test (setup code relies that the
     # changes it made will be seen by the actual test code). So, we use 'cache=shared'
     # in order to force all the connections to use the shared database.
-    url: "jdbc:sqlite:file:memdb?mode=memory&cache=shared"
+    url: "jdbc:sqlite:file:memdb_${random.uuid}?mode=memory&cache=shared"
```
sergeycherkasovv, line 12: Unique DB name per `TestContext` — prevents lock conflicts between reused contexts with `cache=shared`

mipo256: Am I correct that the theory is that in the Spring Boot's context cache we had the context that held the connection to the SQLite database, and because of that a new context that was supposed to use the same in-memory `memdb` failed, right?

PR #1164, opened 2026-06-05
GH-1163 Fix nested REQUIRES_NEW transactions overwriting parent SQL query

### LINKED ISSUE #1163: [BUG] Fix nested REQUIRES_NEW transactions overwriting parent SQL query
Nested REQUIRES_NEW transactions clear or steal the thread-local query records from the outer transaction, causing data loss and inaccurate SQL metrics for parent entrypoints.

NikitaKirilloff:
close #1163 

## Summary by CodeRabbit

* **New Features**
  * Transaction monitoring now isolates SQL query collection per transaction context, ensuring queries from nested transactions are attributed to the correct entrypoint.

* **Tests**
  * Expanded integration tests verifying nested transaction propagation (REQUIRED → REQUIRES_NEW) are reported as separate monitored entrypoints and JSON results are compared order-insensitively.


mipo256 (approved):
That was a good catch. Good job mate @NikitaKirilloff, thanks!

mipo256:
The only thing I did is that I introduced a separate type for the transactions stack. I think it becomes pretty unreadable when we have 3 nesting levels of generics in the type system. 

See: https://github.com/axelixlabs/axelix/pull/1164/changes#diff-76ce4ad0fc96cc8e737944bf0eb93b49dce26fd5a51d0e5cfbb751af579bb27b

PR #1184, opened 2026-06-08
GH-1121 Develop a policy for accessing secrets in the PRs from external contributors

### LINKED ISSUE #1121: [Feature] Develop a policy for accessing secrets in the PRs from external contributors
Secrets seem to be not loaded for external contributors when they make PRs. That appeared to be obvious [in here](https://github.com/axelixlabs/axelix/pull/1087), for example.

We need to develop a policy here.

NikitaKirilloff:
close #1121 

## Summary by CodeRabbit

* **Chores**
  * Refined CI/CD triggers to better target backend changes and exclude front-end-only edits.
  * Converted broad ignore filters to explicit allowlists for deployment runs, reducing unnecessary pipeline executions.
  * Simplified Gradle cache encryption handling by correcting the cache-key input and removing the prior secret-loading step from automated builds.

### DIFF
--- a/.github/actions/build-master/action.yaml
+++ b/.github/actions/build-master/action.yaml
@@ -6,8 +6,7 @@ inputs:
     default: "false"
   gradle-cache-encryption-key:
     description: Gradle configuration cache encryption key
-    required: false
-    default: ""
+    required: true
 
 runs:
   using: composite
--- a/.github/actions/build-master/action.yaml
+++ b/.github/actions/build-master/action.yaml
@@ -4,9 +4,7 @@ inputs:
     description: Whether to deploy the built JAR into Nexus or not
     required: false
     default: "false"
-  gradle-cache-enryption-key:
-    description: Gradle cache encryption key
-    required: true
+
 runs:
   using: composite
   steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/backend_pull_requests.yaml
```diff
@@ -2,11 +2,28 @@ name: Backend Pull Requests
...
+      - 'settings.gradle.kts'
+      - 'gradle.properties'
+      - 'gradlew'
+      - 'gradlew.bat'
+
+      # Code quality, formatting, and license configurations
+      - 'pmd.ruleset.xml'
+      - '.editorconfig'
+      - 'LICENSE'
+      - 'LICENSE_HEADER'
```
NikitaKirilloff, line 26: Added these to paths because Spotless validation relies directly on .editorconfig and license headers during ./gradlew build. If these files change, we must trigger the workflow to ensure formatting checks pass on the server.

.github/workflows/backend_pull_requests.yaml
```diff
@@ -2,11 +2,28 @@ name: Backend Pull Requests
 
 on:
   pull_request:
+    branches: [ master ]
     paths:
+      # Build logic, and application modules
       - 'master/**'
+      - '!master/front-end/**'
```
NikitaKirilloff, line 9: Only master backend, no master/frontend

.github/actions/build-master/action.yaml
```diff
@@ -4,9 +4,7 @@ inputs:
     description: Whether to deploy the built JAR into Nexus or not
     required: false
     default: "false"
-  gradle-cache-enryption-key:
-    description: Gradle cache encryption key
-    required: true
+
```
mipo256, line 10: Let's stick to the paradigm of controlling the behavior clearly from outside for custom GitHub actions. That means the custom GitHub action should not just reference something like `${{ .env }}` inside itself, since this is an implicit dependency of the custom GitHub Action, it is not explicitly declared in the yaml, and this is bad, since in order to understand that there is in fact such dependency - I must scan the source code myself,

mipo256 (changes requested):
The direction is generally good, bu I would fix one thing before merging. I will do that myself

PR #1198, opened 2026-06-10
GH-1074 Added a restriction preventing user deletion under certain conditions

### LINKED ISSUE #1074: [BUG] Articulate the exact behavior of OIDC user being deleted on the confirmation modal
When the user is either not does not have the `UserOrigin` equal to `LOCAL` , or the `LOCAL` auth options is not enabled at all, we should not allow the delteion of the user.

sergeycherkasovv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Deletion now only removes locally created users; attempts to delete external or non-existent users return HTTP 400 instead of silently succeeding.
* **Tests**
  * Added integration tests covering successful local deletions and verifying 400 responses while preserving non-local or missing user records.

* **Changes**
  - `UserRepository` — `deleteById` renamed to `deleteByIdToLocalUser` to reflect the `LOCAL`-only semantics.
  - `UserService` / `DatabaseUserService` — `deleteById` renamed to `deleteByIdToLocalUser` to reflect the `LOCAL`-only semantics. **Javadoc** updated to state that
  non-existent and non-`LOCAL` (`OIDC`) users are a no-op.
  - Condition 2 already enforced: the entire `UserManagementApi` controller is gated by a class-level
  `@ConditionalOnProperty(prefix="axelix.master.auth.options.local", name="enabled", havingValue="true")`, so the `DELETE` endpoint is not exposed (404) when
  local auth is off. No code change required.

### DIFF
--- a/master/src/main/java/com/axelixlabs/axelix/master/repository/UserRepository.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/repository/UserRepository.java
@@ -60,4 +60,8 @@ public interface UserRepository extends ListCrudRepository<UserEntity, String> {
             @Param("password") @Nullable String password,
             @Param("roles") UserEntity.Roles roles,
             @Param("lastLoginAt") @Nullable Instant lastLoginAt);
+
+    @Modifying
+    @Query("DELETE FROM users WHERE id = :id AND user_origin = 'LOCAL'")
+    int deleteByIdToLocalUser(@Param("id") String id);
 }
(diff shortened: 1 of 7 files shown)


master/src/main/java/com/axelixlabs/axelix/master/repository/UserRepository.java
```diff
@@ -60,4 +60,8 @@ void updateUserPatch(
             @Param("password") @Nullable String password,
             @Param("roles") UserEntity.Roles roles,
             @Param("lastLoginAt") @Nullable Instant lastLoginAt);
+
+    @Modifying
+    @Query("DELETE FROM users WHERE id = :id AND user_origin = 'LOCAL'")
+    void deleteByIdToLocalUser(@Param("id") String id);
```
mipo256: Return the count of records deleted. We need to return 400 bad request in case there was an attempt to delete either the non existent user or non-LOCAL user

mipo256 (changes requested):
see my comments

master/src/main/java/com/axelixlabs/axelix/master/repository/UserRepository.java
sergeycherkasovv: done)

mipo256:
You know, I actually think that we may need to be able to delete the user that came from OIDC. Because, for example, in Keycloak or any other SSO OIDC provider the user got removed (e.g. because the person no longer works in a company), and we will retain the record about this user in our database. And there is no way to get rid of it. 

I think that if the end-user that operates on the UI may want to manually delete the user. I think, there is no real problem here. We may later would want to display some sort of the modal that outlines the behavior: deleting the OIDC user from Axelix db does not prevent this user to login (unlike the LOCAL origin case), but in general, we should allow for deleting the OIDC user.

PR #1209, opened 2026-06-10
GH-1202 Remove Logger Converters

### LINKED ISSUE #1202: [Refactoring] Remove Logger Converters

NikitaKirilloff:
close #1202

## Summary by CodeRabbit

* **New Features**
  * Added support for temporary log-level changes with configurable time-to-live (TTL) expiration

* **Refactor**
  * Updated logger management endpoint response structure to use ordered lists for loggers and groups instead of map-based format


master/src/main/java/com/axelixlabs/axelix/master/api/external/request/deserilize/StateExportComponentDeserializer.java
```diff
@@ -64,7 +64,7 @@ private static List<StateComponentSettings> parseComponents(JsonParser p, JsonNo
 
         for (JsonNode childNode : componentsNode) {
             var stateComponentAsText =
-                    childNode.get(StateComponentSettings.COMPONENT).asText();
```
NikitaKirilloff, line 67: `.asText()` is deprecated

master/src/test/java/com/axelixlabs/axelix/master/filter/auth/AbstractMcpAuthorizationFilterTest.java
```diff
@@ -117,7 +117,7 @@ static void stopMockWebServer() throws IOException {
     void setUpRestTemplate() {
         instanceRepository.deleteAll();
         userRepository.deleteAll();
-        this.restTemplate = new TestRestTemplate(new RestTemplateBuilder().rootUri("http://localhost:" + port));
```
NikitaKirilloff, line 120: `rootUril()` is deprecated

common/api/src/main/java/com/axelixlabs/axelix/common/api/loggers/GroupLoggerProfile.java
```diff
@@ -28,48 +28,57 @@
  *
  * @author Sergey Cherkasov
  */
-public final class LoggerGroup {
+public final class GroupLoggerProfile {
```
mipo256:
```suggestion
public final class LoggersGroupProfile {
```

There is not entity called "GroupLogger", there is a "Loggers Group"

common/api/src/main/java/com/axelixlabs/axelix/common/api/loggers/LoggersGroupProfile.java
```diff
@@ -28,48 +28,57 @@
  *
  * @author Sergey Cherkasov
  */
-public final class LoggerGroup {
```
NikitaKirilloff, line 31: I renamed it because it  clashes with `org.springframework.boot.logging.LoggerGroup`

common/api/src/main/java/com/axelixlabs/axelix/common/api/loggers/SingleLoggerProfile.java
```diff
@@ -0,0 +1,139 @@
...
+     * The initial logging level before any temporary modifications were applied.
+     */
+    private final String originalLevel;
+
+    /**
+     * The timestamp indicating when the temporary logging level was activated.
+     * Returns {@code null} if the current level is permanent.
+     */
+    @Nullable
+    private final String temporaryLevelAppliedAt;
```
mipo256, line 61: I would go for unix seconds from epoch and `Long` as the data type. Let's not use raw Strings for dates, that is the recepie for disaster. Using `java.time` types like `Instant` or `OffsetDateTime` is also better than just raw `String`, as they provide a nominal type that explicitly signifies that this is in fact a "Date-Time" something. But they also pose a question of format, like what is the foamtting of those dates etc.

So, just using `Long` is, I guess, the trade-off that is worth having. It is not exactly the date-time type, but we know exactly how we're going to parse it on the receiving end (in master)

mipo256 (changes requested):
Generally very good, a couple of issues needs to be addressed still

common/api/src/main/java/com/axelixlabs/axelix/common/api/loggers/SingleLoggerProfile.java
mipo256: We decided to leave it as it is, since later we're anyway are going to migrate some schema-like files that describe the JSON models. that master exchanges with the starter.

PR #1212, opened 2026-06-10
GH-1050 Remove the clearing stats api for the transactional data

### LINKED ISSUE #1050: [Feature] Remove the clearing stats api for the transactional data

sergeycherkasovv:
## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the ability to clear transaction statistics via the monitoring endpoint. Transaction monitoring is now read-only, supporting statistics retrieval only.

* **Tests**
  * Updated tests to align with the removal of statistics clearing operations.

### DIFF
--- a/sbs/axelix-spring-boot-2-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/transactions/TransactionMonitoringEndpointTest.java
+++ b/sbs/axelix-spring-boot-2-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/transactions/TransactionMonitoringEndpointTest.java
@@ -45,7 +45,6 @@ import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
-import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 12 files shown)


sbs/axelix-spring-boot-2-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/transactions/TransactionMonitoringEndpointTest.java
```diff
@@ -92,7 +91,7 @@ class TransactionMonitoringEndpointTest {
 
     @BeforeEach
     void cleanUp() {
-        transactionStatsCollector.clearAllStats();
+        transactionStatsCollector.getAllStats().clear();
```
sergeycherkasovv: Our method `DefaultTransactionStatsCollector.getAllStats` returns a mutable map. I think I understand why—so that we don’t create copies of the map on every call to the transactions endpoint—but this also makes it a fragile spot. If this behavior is relied on somewhere in the future, it can break the state, as I did in this test. 
What do you think about this, @mipo256 ?

mipo256: You are correct, and I agree with you.

mipo256 (approved):
Overall good, but the nuance with `clear()` and mutation of internal state is an important one

PR #1218, opened 2026-06-12
GH-1216: Added Gradle plugin for Axelix

### LINKED ISSUE #1216: [Feature] Creating Gradle plugin for Axelix

chrshnv:
Closes #1216

## What

Adds the Axelix Gradle plugin under `plugins/axelix-gradle-plugin` with the id `com.axelixlabs.axelix`. When applied to a project (together with the `java` plugin), it:

- adds `digital.pragmatech.testing:spring-test-profiler:0.1.2` to the test runtime classpath;
- generates `META-INF/spring.factories` (into `build/generated/axelix`, wired onto the test runtime classpath) registering `SpringTestProfilerListener` and `ContextDiagnosticApplicationInitializer`, as described in the [profiler docs](https://github.com/PragmaTech-GmbH/spring-test-profiler#2-activate-the-profiler);
- counteracts the Thymeleaf downgrade caused by the Spring Boot 2 BOM when applied via `io.spring.dependency-management` (the profiler needs Thymeleaf 3.1+ to render its report; the rule is scoped to `testRuntimeClasspath` and is a no-op on Boot 3/4 and BOM-less builds).
- copies the profiler HTML report onto the application classpath during `build`: the `copyAxelixTestProfilerReport` task runs after tests pass and copies `build/spring-test-profiler/` into `build/resources/main/spring-test-profiler/`; jar tasks are ordered after the copy so the report deterministically ends up inside the packaged jar.

The dependency is added to `testRuntimeOnly` rather than `testImplementation`: the profiler is activated purely via `spring.factories`, so nothing in the consumer project compiles against it.

## Compatibility

Supports Gradle 4.0 through 9.x. The plugin is compiled to Java 8 bytecode and restricted to APIs with identical signatures in Gradle 4.0 and 9.5.1 (verified via `javap`); notable consequences: eager `tasks.create`, no Provider/Property APIs, no sourceSets access (conventions were removed in Gradle 9), and `TaskInputs.properties(Map)` instead of `property(String, Object)` (its return type changed in 4.3).

## Testing

TestKit functional tests (9 in total, all green):

- both Gradle extremes, 4.0 (daemon forked on a Liberica JDK 8) and 9.5.1: dependency lands on the test runtime classpath, `spring.factories` content is byte-exact, and the resource is visible to tests through the classloader;
- empty Spring Boot 2.7.18 app with `io.spring.dependency-management` (Gradle 8.14) and empty Spring Boot 4.0.7 app (Gradle 9.5.1): after `test`, the profiler report exists at `build/spring-test-profiler/latest.html`. The Boot 2 test fails without the Thymeleaf rule.
- report copy: `build` on the Boot 4 app puts the report at `build/resources/main/spring-test-profiler/latest.html` and packages it into the jar as `spring-test-profiler/latest.html`; `build` on a plain Java project without a report (Gradle 4.0 and 9.5.1) succeeds with a no-op copy.

Also verified manually against a realistic standalone Boot 2.7.18 web app (controller/service, `@SpringBootTest` + `@WebMvcTest`, Boot Gradle plugin + dependency-management) consuming this plugin via composite build: all tests pass and the HTML report is generated.

## Release Notes

* **New Features**
  * Added the Axelix Gradle plugin for Spring Test Profiler integration in Spring Boot projects.
  * Automatically manages required test dependencies (including ensuring a minimum Thymeleaf version when needed).
  * Generates the required `spring.factories` entry and packages the profiler HTML report into your build output.
  * Works across Gradle 4.0 through 9.5.1.

* **Tests**
  * Added functional end-to-end checks verifying generation, dependency behavior, and report packaging.

* **Chores**
  * Updated release metadata/locking information.

### DIFF
--- a/apm.lock.yaml
+++ b/apm.lock.yaml
@@ -1,40 +1,69 @@
 lockfile_version: '1'
-generated_at: '2026-06-09T10:05:07.241939+00:00'
-apm_version: 0.14.1
+generated_at: '2026-06-14T09:23:13.342367+00:00'
+apm_version: 0.20.0
 dependencies:
 - repo_url: _local/backlog-refiner
   package_type: claude_skill
   deployed_files:
   - .agents/skills/backlog-refiner
+  - .agents/skills/backlog-refiner/SKILL.md
+  - .agents/skills/backlog-refiner/evals/evals.json
   - .claude/skills/backlog-refiner
+  - .claude/skills/backlog-refiner/SKILL.md
+  - .claude/skills/backlog-refiner/evals/evals.json
+  deployed_file_hashes:
+    .agents/skills/backlog-refiner/SKILL.md: sha256:84bc6c736d9b64d94622428a5157e061dc223919f50a48fa6a83a75dbad3d068
+    .agents/skills/backlog-refiner/evals/evals.json: sha256:d8a6338ee17ceabd5a2b32f247256d6986058db2a48e1e796febcf86fd255208
+    .claude/skills/backlog-refiner/SKILL.md: sha256:84bc6c736d9b64d94622428a5157e061dc223919f50a48fa6a83a75dbad3d068
+    .claude/skills/backlog-refiner/evals/evals.json: sha256:d8a6338ee17ceabd5a2b32f247256d6986058db2a48e1e796febcf86fd255208
   source: local
   local_path: ./.agent_skills/backlog-refiner
 - repo_url: _local/docs-writer
   package_type: claude_skill
   deployed_files:
   - .agents/skills/docs-writer
+  - .agents/skills/docs-writer/SKILL.md
   - .claude/skills/docs-writer
+  - .claude/skills/docs-writer/SKILL.md
+  deployed_file_hashes:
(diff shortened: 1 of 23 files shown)


mipo256 (changes requested):
Thanks, @chrshnv !

Although, there are a couple of things that I think that make this PR a bit less than what it otherwise can be.

1. First of all, inlining Gradle configuration or test source files as regular strings is highly unreadable. We at least need to extract them into separate files

2. The plugin will have the same versioning scheme as the main Axelix project (it is part of the main Axelix project in fact). And the current version of Axelix is already defined in `gradle.properties`. Ideally, we would not like to duplicate it in this local `gradle.properties`.

Now, the reason we have to do it is the Gradle composite build. This is the thing that I have not foreseen with a composite build. I think we should try to incorporate the Gradle plugin as the Gradle sub-project to our monorepo, i.e. we will not need separate `gradle.properties` and `settings.gradle.kts`, WDYT @chrshnv ?

3. Apart from that, I think, that we should try not to include Spring Boot Test Profiler and Thymeleaf in case they are already present in the classpath for tests - we can cause real problems in user's build if we include these in such case.

Apart from that, the PR is more or less fine, it needs minor polishing from my point of view, but that is a secondary order concern.

PR #1226, opened 2026-06-15
GH-1139 Fix flaky transaction monitoring timestamp nesting invariant

### LINKED ISSUE #1139: [BUG] Flaky transaction monitoring test: query end timestamp may exceed execution end timestamp by 1 ms
## Description

`TransactionMonitoringEndpointTest.shouldReturnsStatsAfterTransactionExecution()` is flaky in the Spring Boot 2 starter test suite.

During a full build, the test failed because the recorded query `endTimestampMs` was 1 ms greater than the parent execution `endTimestampMs`. The test currently assumes that every query timestamp is strictly contained within the execution timestamp range.

## Failed command

```bash
./gradlew clean build
```

Failed task:

```text
:sbs:axelix-spring-boot-2-starter:test
```

## Failure

```text
TransactionMonitoringEndpointTest > shouldReturnsStatsAfterTransactionExecution() FAILED
    java.lang.AssertionError at TransactionMonitoringEndpointTest.java:134

Expecting actual:
  1780345191934L
to be between:
  [1780345191933L, 1780345191933L]
```

The failing assertion is:

```java
assertThat(queryEndTimestampMs).isBetween(executionStartTimestampMs, executionEndTimestampMs);
```

## Observed behavior

The query timestamps may be recorded like this:

```text
executionStartTimestampMs = 1780345191933
executionEndTimestampMs   = 1780345191933
queryEndTimestampMs       = 1780345191934
```

So the query end timestamp is just outside the execution interval by 1 ms.

## Expected behavior

Either:

1. transaction execution timestamps should always include all tracked query timestamps, or
2. the test should allow a small timestamp tolerance if the implementation only guarantees approximate millisecond-level timing.

## Possible fix

For the test-level fix, replace the strict upper-bound assertion for `queryEndTimestampMs` with an AssertJ `isCloseTo(...)` assertion using a small tolerance, for example 1 ms:

```java
assertThat(queryEndTimestampMs).isCloseTo(executionEndTimestampMs, within(1L));
```

sergeycherkasovv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction execution timing so the transaction end time and published metrics reflect the latest completion time of any nested SQL query within the transaction.
* **New Features**
  * Enhanced transaction timing data by adding an end-timestamp accessor on transaction records (with duration derived from start/end).
* **Documentation**
  * Clarified that returned query history is chronologically ordered to better reflect the sequence of transaction activity.

### DIFF
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/DefaultTransactionMonitoringService.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/DefaultTransactionMonitoringService.java
@@ -107,16 +107,9 @@ public class DefaultTransactionMonitoringService implements TransactionMonitorin
     private TransactionExecution convertToTransactionExecution(TransactionRecord record) {
         List<Query> queries = convertToQueries(record.getQueries());
         long startTimestamp = record.getStartTimestampMs();
-        long endTimestamp = startTimestamp + record.getDurationMs();
+        long endTimestamp = record.getEndTimestampMs();
 
-        // The execution end and each query end are synthesized from currentTimeMillis() plus a
-        // nanoTime-based duration truncated to whole milliseconds. A query is always nested inside
-        // the execution in real time, so make the reported execution end cover the latest query end
-        // to keep that invariant after millisecond rounding.
-        long maxQueryEndTimestamp =
-                queries.stream().mapToLong(Query::getEndTimestampMs).max().orElse(endTimestamp);
-
-        return new TransactionExecution(startTimestamp, Math.max(endTimestamp, maxQueryEndTimestamp), queries);
+        return new TransactionExecution(startTimestamp, endTimestamp, queries);
     }
 
     private List<Query> convertToQueries(List<SqlQueryRecord> queriesRecords) {
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/DefaultTransactionMonitoringService.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/DefaultTransactionMonitoringService.java
@@ -108,7 +108,15 @@ public class DefaultTransactionMonitoringService implements TransactionMonitorin
         List<Query> queries = convertToQueries(record.getQueries());
         long startTimestamp = record.getStartTimestampMs();
         long endTimestamp = startTimestamp + record.getDurationMs();
-        return new TransactionExecution(startTimestamp, endTimestamp, queries);
+
+        // The execution end and each query end are synthesized from currentTimeMillis() plus a
+        // nanoTime-based duration truncated to whole milliseconds. A query is always nested inside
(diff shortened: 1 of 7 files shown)


sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/transactions/DefaultTransactionMonitoringService.java
```diff
@@ -108,7 +108,15 @@ private TransactionExecution convertToTransactionExecution(TransactionRecord rec
...
-        return new TransactionExecution(startTimestamp, endTimestamp, queries);
+
+        // The execution end and each query end are synthesized from currentTimeMillis() plus a
+        // nanoTime-based duration truncated to whole milliseconds. A query is always nested inside
+        // the execution in real time, so make the reported execution end cover the latest query end
+        // to keep that invariant after millisecond rounding.
+        long maxQueryEndTimestamp =
+                queries.stream().mapToLong(Query::getEndTimestampMs).max().orElse(endTimestamp);
+
+        return new TransactionExecution(startTimestamp, Math.max(endTimestamp, maxQueryEndTimestamp), queries);
```
sergeycherkasovv: Thanks @CoolXaleX for spotting this. It seems to me the root cause isn't in the tests but in how we derive the end timestamps: both the **execution** end and each **query **end are synthesized as `currentTimeMillis()` (**ms**) plus a `nanoTime`-based duration truncated to whole **milliseconds**. A query is always nested inside  its execution in real time, but this mix of clocks plus ms rounding can push the query end up to 1 ms past the execution end.

At scale this matters: with many transactions, a fraction of the short ones will show queries finishing outside their transaction boundary — leading to misleading data, broken timelines on the graphs, and wrong conclusions.

The fix takes the **execution** end as the later of its own computed end and the latest **query** end: if a query appears to finish after the transaction (the ≤1 **ms** rounding artifact), we extend the reported transaction end to that query's end; otherwise we keep the transaction's own end. So it seems to me we don't lose any timing accuracy

What do you think, @mipo256?

mipo256: That is a brilliant observation, and I thank you for that @sergeycherkasovv. I agree

mipo256: Although, I think the current implementation is a bit off, since we should not do this logic inside the `DefaultTransactionMonitoringService`, rather, we should do it when we record the transaction itself, since the current solution has flaws, in particular:

1. Does not anyhow affect the metrics 

2. And it basically defeats the purpose of `durationMs` inside `SqlQueryRecord`, since in this case, we're kind of saying: "This is the transaction duration, but not all the time, sometimes, we calculte this duration based on..." - and that is just breaking the encapsulation of the complexity. The code that is external to `SqlQueryRecord` should just rely on the fact that `durationMs` is the duration of the transaction, **it always is.**

mipo256 (approved):

PR #1227, opened 2026-06-16
test(sbs3): share one Spring context across beans integration tests

chrshnv:
## What

The `...core.beans` tests in `sbs/axelix-spring-boot-3-starter` each built their own Spring `TestContext`. This makes the two **non-endpoint** beans tests share a single cached context:

- `DefaultBeanMetaInfoExtractorTest`
- `QualifiersPersistencePostProcessorTest`

They now extend a new abstract `AbstractSharedBeansContextTest`. The base pins `Main` via `@SpringBootTest(classes = Main.class)` and contributes the union of both tests' fixtures via `@Import`, so Spring reuses one `ApplicationContext` for both.

`AxelixBeansEndpointTest` (the endpoint test) is intentionally left on its own context.

## Why `Main` is pinned in `classes`

Leaving the configuration to Spring Boot's auto-detection produced divergent, non-shareable merged configurations across the two subclasses (and, in one attempt, dropped the auto-detected `Main` so the embedded `DataSource` was missing). Pinning `Main` explicitly via the `classes` attribute makes both subclasses resolve an identical, deterministic context — which is what lets them share the cache — while keeping auto-configuration (e.g. the embedded `DataSource`) active. `QualifiersPersistencePostProcessor` is not imported separately; it is already provided by `DefaultBeanAnalyzerTestConfig` as a static `@Bean`.

## Verification

- `./gradlew :sbs:axelix-spring-boot-3-starter:build` — green (tests + Spotless + PMD).
- spring-test-profiler report confirms **one shared context** for `DefaultBeanMetaInfoExtractorTest` + `QualifiersPersistencePostProcessorTest` (16 test methods, one context), and a **separate** context for `AxelixBeansEndpointTest`.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AbstractSharedBeansContextTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AbstractSharedBeansContextTest.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


mipo256 (approved):
Do not forget to add yourself as the author, I will od that myself

PR #1228, opened 2026-06-16
test(sbs3): share one Spring context across env integration tests

chrshnv:
## What & why

Each non-endpoint test under `sbs/axelix-spring-boot-3-starter/.../core/env/` declared its own `@SpringBootTest` configuration, so every class — and each `@Nested` scenario — built a separate `ApplicationContext`. This consolidates them onto a single shared, cached context.

## How

- **`AbstractEnvSharedContextTest`** (new abstract base): one `@SpringBootTest(args=…)` + the union of the tests' `@TestPropertySource` properties + `@EnableConfigurationProperties` + `@Import(EnvSharedTestConfig)`. The four non-endpoint env tests now extend it, so Spring caches the context once and reuses it.
- **`EnvSharedTestConfig`** (new shared `@TestConfiguration`): the union of beans needed. The differing sanitization behaviour is exposed as **two separately-named** `SmartSanitizingFunction` beans (`sanitizeAllSmartSanitizingFunction` / `explicitSmartSanitizingFunction`), each backing a named `EnvironmentService`, so a single context satisfies every scenario without ambiguous autowiring. The `@Value` sample beans and the `AxelixConfigurationProperties` record were relocated here.
- The Actuator endpoint test (`AxelixEnvironmentEndpointTest`) is intentionally **left on its own context**.

## Result (spring-test-profiler, env package)

| Metric | Before | After |
|---|---|---|
| Cache Size (contexts) | 6 | **2** |
| Cache Misses | 6 | **2** |
| Cache Hit Rate | 87.5% | **96.2%** |

The shared context holds exactly the 6 env units (`DefaultEnvPropertyEnricherTest`, `DefaultEnvironmentServiceTest` + its 2 `@Nested` scenarios, `DefaultPropertyMetadataExtractorTest`, `ValueInjectionTrackerBeanPostProcessorTest`); `AxelixEnvironmentEndpointTest` stands alone.

## Verification

- Env tests green; module `spotlessCheck` / `pmdMain` / `pmdTest` / Error-Prone+NullAway compile green.
- Behaviour preserved; assertions unchanged except the value-injection tracker's expected bean name (sample beans now registered via `@Bean` in the shared config).

## Summary by CodeRabbit

* **Tests**
  * Refactored integration test infrastructure to consolidate shared configuration and reduce duplication across test classes.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/EnvSharedTestConfig.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/EnvSharedTestConfig.java
@@ -55,6 +55,7 @@ import com.axelixlabs.axelix.sbs.spring.core.configprops.SmartSanitizingFunction
  *
  * @author Nikita Kirillov
  * @author Mikhail Polivakha
+ * @author Artemiy Degtyarev
  */
 @TestConfiguration
 public class EnvSharedTestConfig {
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/EnvSharedTestConfig.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/EnvSharedTestConfig.java
@@ -1,0 +1,237 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.sbs.spring.core.env;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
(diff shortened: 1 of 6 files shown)


sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/ValueInjectionTrackerBeanPostProcessorTest.java
```diff
@@ -17,60 +17,30 @@
...
 
-    @TestConfiguration
-    static class ValueInjectionTrackerBeanPostProcessorTestConfig {
-
-        @Bean
-        public static ValueInjectionTrackerBeanPostProcessor valueInjectionTrackerBeanPostProcessor() {
-            return new ValueInjectionTrackerBeanPostProcessor(new DefaultPropertyNameNormalizer());
-        }
-    }
+    private static final String SAMPLE_BEAN_NAME = "testBeanWithCustomAnnotations";
```
mipo256, line 43: This constant should be transferred to `EnvSharedTestConfig`, and also the corresponding bean should have the qualifier annotation that explicitly pins the `"testBeanWithCustomAnnotations"` qualifier, since in the current case this dependency is implicit. If somebody renames the `@Bean` method from `testBeanWithCustomAnnotations()` to something else - welcome to spend half an hour debugging why the test broke

sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/env/EnvSharedTestConfig.java
```diff
@@ -0,0 +1,238 @@
...
+     * Declared {@code static} so the post-processor is registered early and analyzes every bean in the context
+     * (including the {@code @Value} sample beans below).
+     */
+    @Bean
+    public static ValueInjectionTrackerBeanPostProcessor valueInjectionTrackerBeanPostProcessor() {
+        return new ValueInjectionTrackerBeanPostProcessor(new DefaultPropertyNameNormalizer());
+    }
+
+    @Bean
+    public SmartSanitizingFunction sanitizeAllSmartSanitizingFunction(PropertyNameNormalizer propertyNameNormalizer) {
```
mipo256, line 112: We need to give explicit bean names here as well, for the same reason

mipo256 (approved):
Overall pretty good, I will polish the minor things myself, please read the comments. No action needed, I will polish after the PR will be merged.

Good work @chrshnv

PR #1229, opened 2026-06-16
test(sbs3): share one Spring context across endpoint integration tests

chrshnv:
## What

The 14 actuator endpoint integration tests in `axelix-spring-boot-3-starter` each declared their own `@SpringBootTest` / `@Import` / `@TestPropertySource`, so every class resolved to a distinct Spring `TestContext` cache key and booted a fresh `RANDOM_PORT` web server. This change collapses all 14 onto **one shared `ApplicationContext`**.

## How

- Add `AbstractEndpointIntegrationTest` (the sole `@SpringBootTest` carrier) and `SharedEndpointTestConfiguration` (one aggregate `@Import` of every endpoint's beans + the nested per-test `@TestConfiguration`s — kept nested so assertions reading enclosing class names still work).
- Convert all 14 `*EndpointTest` classes to `extends` the base, removing the per-test context-affecting annotations so they share one identical cache key.
- Move endpoint-only properties into `shared-endpoint-test.properties` (loaded only by the base) so non-endpoint unit tests are not polluted.
- Resolve the conflicts surfaced by merging the contexts:
  - dedupe `libraryInformationProvider` / git provider (metadata vs details), `axelixMetricsPublisher` (caches vs transaction), and `smartSanitizingFunction` + the configprops infrastructure beans (env vs configprops), using `@Primary` and ordering `EnvironmentTestConfig` before `JwtAuthTestConfiguration` for the `@ConditionalOnMissingBean` `securityContextExecutor`.
- Harden shared mutable state:
  - the loggers test re-establishes its baseline log levels in `@BeforeEach` (the JVM-global `LoggingSystem` is re-initialised by other contexts);
  - the transaction test removes only its own meters instead of clearing the shared `MeterRegistry` (which would wipe the metrics test's binder-registered meters).
- Rewrite the env precedence assertions (now system-property vs test-property-file precedence, after dropping `args` / `@DynamicPropertySource`) and bean-scope the env/configprops feed filters, since `axelix.prop.test` is now bound by several `@ConfigurationProperties` beans.

## Verification

`spring-test-profiler` report: context cache entries **34 → 21** — the 14 endpoint test classes now share a single context (1 miss + 13 hits), hit rate 87.6% → 92.4%.

Full module build is green: `spotlessCheck`, `pmdMain`, `pmdTest`, and the complete test suite pass.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AxelixBeansEndpointTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AxelixBeansEndpointTest.java
@@ -42,6 +42,7 @@ import static org.assertj.core.api.InstanceOfAssertFactories.type;
  * Integration tests for {@link AxelixBeansEndpoint}.
  *
  * @author Mikhail Polivakha
+ * @author Artemiy Degtyarev
  */
 public class AxelixBeansEndpointTest extends AbstractEndpointIntegrationTest {
 
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AxelixBeansEndpointTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AxelixBeansEndpointTest.java
@@ -21,26 +21,19 @@ import java.util.function.Supplier;
 
 import org.junit.jupiter.api.Test;
 
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.actuate.autoconfigure.condition.ConditionsReportEndpoint;
-import org.springframework.boot.actuate.beans.BeansEndpoint;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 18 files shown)


mipo256 (changes requested):
I am not exactly sure that this is a good design. Again, the problem is that the `SharedEndpointTestconfiguration` knows about its implementations. It would not be that huge of a deal, like in case of `AbstractEnvSharedContextTest`, since there it is just the comments that simply hint what is being used where. And TBH even this part I am not really sure is the way to go.

Here, we have direct imports from children, which is a non-starter.

The core of the problem that we have is that each endpoint has some configurations that are specific to its testing. We can, of course move all the test configuration classes onto master and this will just be fine, but I am not really sure this is a good solution either. 

Maybe we should explore the ways to merge endpoints tests contexsts into their dedicated infastructure test contexts? In other words, making `AbstractSharedBeansContextTest` creating all the needed infrastructure for beans, including for the `BeansEndpoint`, and so on? I think this approach would be more viable and more clear, WDYT? @chrshnv 

P.S: I also think that in `**ContextTest` classes we should generally not use `@SpringBootTest`, since in this case we're risking scanning a lot of beans that we do not really need, and we still end up with the huge context. 

**I am not asking you to do anything about it right now**. This is another issue. If we do not use `@SpringBooTest` in shared context classses we risk breaking tests. We're going to do that separately. I will explore that problem myself and outline my thoughts in the separte ticket.

PR #1232, opened 2026-06-16
test(sbs3): share one Spring context across scheduled integration tests

chrshnv:
## What

`ScheduledTaskServiceTest` and `ScheduledTasksRegistryTest` in the Spring Boot 3 starter each spun up their **own** Spring `ApplicationContext`, because each imported a different inner `@TestConfiguration` — giving them distinct `MergedContextConfiguration` cache keys. Context startup dominates these tests' runtime, so this is wasted work.

This PR makes the two share a single cached context.

## How

- Add `AbstractScheduledTasksIntegrationTest` — an abstract base carrying the common `@SpringBootTest` + `@Import` of a single **superset** `@TestConfiguration`. Spring's TestContext framework merges inherited class-level annotations, so two subclasses that add no further context configuration produce an identical cache key and reuse one context.
- `ScheduledTaskServiceTest` and `ScheduledTasksRegistryTest` now extend the base; their per-class `@SpringBootTest`/`@Import`/inner configs are removed and task-id constants repointed to the shared config.
- `ScheduledTasksRegistryTest` is read-only; its assertions are made **state- and order-independent** (asserts the registry *contains* the expected task ids and *has at least one* of each task type) so that the context mutated by `ScheduledTaskServiceTest` cannot cause order-dependent failures.
- `AxelixScheduledTasksEndpointTest` is intentionally left on its own context (RANDOM_PORT web environment + property source + JWT auth configuration).

## Verification

- `./gradlew :sbs:axelix-spring-boot-3-starter:test` — green (Spotless/PMD/NullAway included).
- `spring-test-profiler` report confirms `ScheduledTaskServiceTest` + `ScheduledTasksRegistryTest` map to **one** MOCK context, while `AxelixScheduledTasksEndpointTest` keeps a separate (RANDOM_PORT) context.
- Flakiness check: forced `ClassName` ordering so the mutating `ScheduledTaskServiceTest` runs first, then the read-only `ScheduledTasksRegistryTest` against the dirtied context — green across repeated runs and the reverse order.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/AbstractScheduledSharedContextTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/AbstractScheduledSharedContextTest.java
@@ -45,9 +45,7 @@ import org.springframework.scheduling.config.ScheduledTaskRegistrar;
  * <p>The shared context is keyed by the merged context configuration of this class. Because
  * Spring's TestContext framework merges inherited class-level annotations, every concrete
  * subclass that adds no further context configuration produces an identical cache key and
- * therefore reuses the very same cached context. The {@link AxelixScheduledTasksEndpointTest}
- * deliberately does <em>not</em> extend this class: it runs on a real web server with its own
- * property source and security configuration, so it keeps a separate context.
+ * therefore reuses the very same cached context.
  *
  * @since 16.06.2026
  * @author Nikita Kirillov
@@ -55,8 +53,8 @@ import org.springframework.scheduling.config.ScheduledTaskRegistrar;
  * @author Artemiy Degtyarev
  */
 @SpringBootTest
-@Import(AbstractScheduledTasksIntegrationTest.SharedScheduledTasksTestConfiguration.class)
-abstract class AbstractScheduledTasksIntegrationTest {
+@Import(AbstractScheduledSharedContextTest.SharedScheduledTasksTestConfiguration.class)
+abstract class AbstractScheduledSharedContextTest {
 
     static volatile boolean cronFlag = false;
 
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/ScheduledTasksRegistryTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/ScheduledTasksRegistryTest.java
@@ -38,7 +38,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Nikita Kirillov
  * @author Artemiy Degtyarev
  */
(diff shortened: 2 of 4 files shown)


sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/ScheduledTasksRegistryTest.java
```diff
@@ -70,18 +64,20 @@ class ScheduledTasksRegistryTest {
...
                 registeredTasks.stream().map(ManagedScheduledTask::getId).toList();
 
         assertThat(taskIds)
-                .hasSize(4)
                 .allSatisfy(id -> assertThat(id).isNotBlank())
-                .containsExactlyInAnyOrder(CRON_TASK_ID, FIXED_DELAY_TASK_ID, FIXED_RATE_TASK_ID, CUSTOM_TASK_ID);
-
-        assertThat(registeredTasks)
-                .allSatisfy(task -> assertThat(task.getFuture().isCancelled()).isFalse());
+                .contains(
```
mipo256, line 72:
```suggestion
                .containsInAnyOrder(
```

Since the interface api does not guarantee the exact ordering.

sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/scheduled/AbstractScheduledSharedContextTest.java
```diff
@@ -0,0 +1,162 @@
...
+ * @since 16.06.2026
+ * @author Nikita Kirillov
+ * @author Sergey Cherkasov
+ * @author Artemiy Degtyarev
+ */
+@SpringBootTest
+@Import(AbstractScheduledSharedContextTest.SharedScheduledTasksTestConfiguration.class)
+abstract class AbstractScheduledSharedContextTest {
+
+    static volatile boolean cronFlag = false;
```
mipo256, line 59: That is a non-starter.

In case that for any reason we will have two contexts created with `AbstractScheduledSharedContextTest` as the parent, we're going to have a shared mutable state that is going to be changed concurrently under race in both contexts.

It was acceptable to have something like that inside a particular test class (although, I should admit that this is **not** really a good test design), but when it becomes shared - it opens up the door for flaky tests and bugs that are insanly hard to debug.

mipo256 (changes requested):
This pr has some flaws, I am going to adjust them myself, but take a look, please, @chrshnv.

mipo256:
After giving it some thought, I even think that merging these contexts is a mistake.

The reason it is a mistake is because we fundamentally have the following behavior:

1. One test class contains tests that attempt to modify the state of the context, so the context de-facto gets dirty.
2. And the other test just reads the data without attempts to modify anything.

I think that this is a problem. It is a problem because we literally have the shared mutable context in this case, and this fact itself can and probably will lead to some bugs in the future. We're not going to gain a lot from merging this exact two contexts, so I think we should abadon this particular change, at least for now.

CC: @chrshnv

PR #1233, opened 2026-06-16
test(sbs3): share one Spring context across configprops integration tests

chrshnv:
## What

Collapse the `configprops` integration tests in `sbs/axelix-spring-boot-3-starter` onto a **single cached Spring `ApplicationContext`**. Previously the Converter test and both Service scenarios forked three distinct contexts (differing properties, imports, and `SmartSanitizingFunction` beans).

## How

- Add `AbstractConfigPropsIntegrationTest` — one `@SpringBootTest(classes = Main.class)` + shared `@TestPropertySource` + `@Import`, giving every subclass an identical merged context configuration (identical cache key).
- Add `ConfigPropsTestSupportConfiguration` — shared beans plus a single `SharedAxelixConfigurationProperties` binding. The two sanitization behaviors are now two **named** `DefaultConfigurationPropertiesService` beans (`sanitizeAllConfigurationPropertiesService` / `explicitlySanitizedConfigurationPropertiesService`) declared in that shared context.
- `DefaultConfigurationPropertiesServiceTest` selects the right service per `@Nested` scenario via `@Qualifier`.
- `DefaultConfigurationPropertiesConverterTest` reuses the shared context (assertions unchanged).
- `AxelixConfigurationPropertiesEndpointTest` keeps its own `RANDOM_PORT` context; `SmartSanitizingFunctionTest` is a plain unit test and is left untouched.

## Verification

Confirmed via the **spring-test-profiler** report (`build/spring-test-profiler/latest.html`):

- **Cache Size: 2** (down from 4 distinct configprops contexts) — 28 cache hits, 93.3% hit rate.
- `context-1` shared by `DefaultConfigurationPropertiesConverterTest` + both `DefaultConfigurationPropertiesServiceTest` `@Nested` scenarios.
- `context-0` isolated for `AxelixConfigurationPropertiesEndpointTest`.

Tests, Spotless, and PMD all pass for the module.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/DefaultConfigurationPropertiesServiceTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/DefaultConfigurationPropertiesServiceTest.java
@@ -37,6 +37,8 @@ import com.axelixlabs.axelix.common.auth.core.SecurityContext;
 import com.axelixlabs.axelix.common.auth.core.User;
 import com.axelixlabs.axelix.sbs.spring.core.auth.ThreadLocalSecurityContextExecutor;
 
+import static com.axelixlabs.axelix.sbs.spring.core.configprops.ConfigPropsTestSupportConfiguration.EXPLICITLY_SANITIZED_CONFIGURATION_PROPERTIES_SERVICE;
+import static com.axelixlabs.axelix.sbs.spring.core.configprops.ConfigPropsTestSupportConfiguration.SANITIZE_ALL_CONFIGURATION_PROPERTIES_SERVICE;
 import static com.axelixlabs.axelix.sbs.spring.core.utils.UserUtils.createUserWithAuthorities;
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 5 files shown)


sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/DefaultConfigurationPropertiesServiceTest.java
```diff
@@ -111,30 +97,13 @@ void shouldReturnUnsanitizedConfigurationProperties_whenUserHasRequiredAuthority
...
-                "axelix.prop.test.tags.forSanitization=toBeSanitized",
-                "axelix.prop.test.tags.FOR_SANITIZATION=toBeSanitized"
-            })
     @Nested
-    @EnableConfigurationProperties(AxelixConfigurationProperties.class)
-    @Import(TestConfigWithExplicitSanitizationProperties.class)
     class WithExplicitSanitizationProperties {
 
         @Autowired
+        @Qualifier("explicitlySanitizedConfigurationPropertiesService")
```
mipo256:
```suggestion
        @Qualifier(SOME_PUBLIC_STATIC_IMMUTABLE_FIELD_TO_BE_REFERENCED)
```

Again, we should refrain from relying on the default qualifier in Spring Framework. If we do that, the name of the `@Bean` will get the second _implicit_ meaning - it is not just the name of the method that produces the bean, but now, the bean name becomes important. 

Generally, bean **name must not be that important**, because bean names are not widely considered to be part of the contract of the given bean. But even in cases where we want to rely on bean name (**in tests that is okay**), we should make this dependency explicit, not implicit.

mipo256 (approved):
Overall we're pretty good.

PR #1236, opened 2026-06-16
GH-1216: Contribute test dependency and generate spring.factories (2/3)

### LINKED ISSUE #1216: [Feature] Creating Gradle plugin for Axelix

chrshnv:
Part **2/3** of splitting #1218 (closes #1216). Builds on #1235.

## What

Wires the Spring Test Profiler activation into the plugin:

- `TestDependencyContributor` adds `digital.pragmatech.testing:spring-test-profiler:0.1.2`
  to `testRuntimeOnly` when absent, and contributes/bumps Thymeleaf to `3.1.3`
  (the profiler's HTML report fails to render on older Thymeleaf), leaving a
  project's own declarations untouched.
- `SpringFactoriesGenerator` writes `META-INF/spring.factories` registering the
  profiler's `TestExecutionListener` and `ApplicationContextInitializer` onto the
  test runtime classpath via the `generateAxelixSpringFactories` task.
- `BuildDirCompat` resolves the build directory across Gradle 4.0 through 9.x.

The dependency is added to `testRuntimeOnly` (not `testImplementation`): the
profiler is activated purely via `spring.factories`, so nothing compiles against
it.

## Testing

TestKit functional tests on both Gradle extremes (4.0 and 9.5.1): the dependency
lands on the test runtime classpath, `spring.factories` content is byte-exact,
the resource is visible to tests through the classloader, pre-declared
profiler/Thymeleaf versions are left untouched, and an outdated Thymeleaf is
bumped.

## Splitting note

Logically stacked on #1235; since these target this repo from a fork, the diff
shown here includes the 1/3 commit. Review the `GH-1216: Contribute test
dependency...` commit.

### DIFF
--- a/plugins/axelix-gradle-plugin/build.gradle.kts
+++ b/plugins/axelix-gradle-plugin/build.gradle.kts
@@ -1,0 +1,33 @@
+plugins {
+    `java-gradle-plugin`
+}
+
+repositories {
+    mavenCentral()
+}
+
+// The plugin must run inside Gradle 4.0 daemons, which require JDK 8.
+tasks.compileJava {
+    options.release = 8
+}
+
+gradlePlugin {
+    plugins {
+        create("axelix") {
+            id = "com.axelixlabs.axelix"
+            implementationClass = "com.axelixlabs.gradle.plugin.AxelixGradlePlugin"
+        }
+    }
+}
+
+dependencies {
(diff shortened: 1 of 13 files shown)


chrshnv:
Superseded by splitting this into two PRs: #1238 (add the test dependency, 2/4) and #1239 (generate spring.factories, 3/4). The branch is left in place but this PR is closed in favor of the finer-grained pair.

PR #1238, opened 2026-06-16
GH-1216: Contribute the Spring Test Profiler test dependency (2/4)

### LINKED ISSUE #1216: [Feature] Creating Gradle plugin for Axelix

chrshnv:
Part **2/4** of splitting #1218 (closes #1216). Builds on #1235.

## What

`TestDependencyContributor` adds `digital.pragmatech.testing:spring-test-profiler:0.1.2`
to `testRuntimeOnly` when absent, and contributes/bumps Thymeleaf to `3.1.3`
(the profiler's HTML report fails to render on older Thymeleaf), leaving a
project's own declarations untouched. The contribution is deferred to
`afterEvaluate` so the project's own dependency declarations are visible.

The dependency is added to `testRuntimeOnly` (not `testImplementation`): the
profiler is activated purely via `spring.factories` (added in 3/4), so nothing
compiles against it.

## Testing

TestKit functional tests on both Gradle extremes (4.0 and 9.5.1): the profiler
lands on the test runtime classpath, a pre-declared profiler/Thymeleaf version
is left untouched, and an outdated Thymeleaf is bumped.

## Splitting note

This further splits the old combined PR (#1236, now closed) into dependency
(2/4) and `spring.factories` (3/4). Logically stacked on #1235; since these
target this repo from a fork, the diff shown here includes the 1/4 commit.
Review the `GH-1216: Contribute the Spring Test Profiler test dependency` commit.

## Summary by CodeRabbit

* **New Features**
  * The Gradle plugin now adds the required test profiler and a compatible Thymeleaf dependency to the test runtime classpath when they’re missing.
  * Adds semantic version “at least” checking to support compatibility decisions.
* **Bug Fixes**
  * Improved plugin behavior to apply dependency configuration after the Java plugin is available, ensuring reliable setup.
  * Avoids changing dependencies when an incompatible Thymeleaf version is already declared.
* **Tests**
  * Added TestKit functional coverage across multiple Gradle versions and compatibility scenarios.
* **Chores**
  * Updated build compilation to target Java 11 and expanded test execution for modern/legacy Gradle versions.


plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/AxelixGradlePlugin.java
```diff
@@ -28,7 +28,7 @@
  * wiring, no sourceSets/conventions access ({@code JavaPluginConvention} was removed in Gradle 9,
  * {@code JavaPluginExtension.getSourceSets()} was only added in 7.1).
  *
- * <p>The plugin defers all work until the {@code java} plugin is applied, because the
+ * <p>The plugin is a no-op until the {@code java} plugin is applied, because the
  * {@code testRuntimeOnly} configuration only exists once it is.
```
mipo256: Add yourself as the author, please @chrshnv

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
```diff
@@ -0,0 +1,152 @@
...
+ * Contributes the Spring Test Profiler and Thymeleaf to the project's test classpath when they are
+ * missing, leaving a project's own declarations untouched.
+ */
+final class TestDependencyContributor {
+
+    static final String PROFILER_DEPENDENCY = "digital.pragmatech.testing:spring-test-profiler:0.1.2";
+
+    static final String PROFILER_GROUP = "digital.pragmatech.testing";
+
+    static final String PROFILER_NAME = "spring-test-profiler";
```
mipo256:
```suggestion
// spring boot test profiler
    static final String PROFILER_GROUP = "digital.pragmatech.testing";
    static final String PROFILER_NAME = "spring-test-profiler";
    static final String PROFILER_VERSION = "0.1.2";
    static final String PROFILER_DEPENDENCY = PROFILER_GROUP + ":" +  PROFILER_NAME + ":" + PROFILER_VERSION;
```

Let's do the same with thymeleaf

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
```diff
@@ -0,0 +1,152 @@
...
+        return findDeclaredOnTestClasspath(project, group, name) != null;
+    }
+
+    /**
+     * Returns the dependency with the given group and name declared on the test runtime classpath,
+     * including dependencies inherited from extended configurations (e.g. an {@code implementation}
+     * dependency that propagates to {@code testRuntimeClasspath}), or {@code null} if none is
+     * declared. Only declared dependencies are inspected; the configuration is not resolved.
+     */
+    private static Dependency findDeclaredOnTestClasspath(
```
mipo256:
```suggestion
    private static Dependency findInTestClasspath(
```

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
```diff
@@ -0,0 +1,152 @@
...
+
+    private static boolean isOnTestClasspath(final Project project, final String group, final String name) {
+        return findDeclaredOnTestClasspath(project, group, name) != null;
+    }
+
+    /**
+     * Returns the dependency with the given group and name declared on the test runtime classpath,
+     * including dependencies inherited from extended configurations (e.g. an {@code implementation}
+     * dependency that propagates to {@code testRuntimeClasspath}), or {@code null} if none is
+     * declared. Only declared dependencies are inspected; the configuration is not resolved.
```
mipo256: Does that account for transitive dependencies? Please, make sure it does. [Becuase thymleaf os also quite often used as a transitive dependency](https://mvnrepository.com/artifact/org.thymeleaf/thymeleaf/used-by). We need to be carefull here, since we're risking to break the classpath if we do not traverse the transitive dependencies as well

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
```diff
@@ -0,0 +1,152 @@
...
+        String numericVersion = leadingNumericVersion(version);
+        if (numericVersion == null) {
+            return false;
+        }
+        return GradleVersion.version(numericVersion).getBaseVersion().compareTo(MIN_THYMELEAF_VERSION) < 0;
+    }
+
+    /**
+     * Extracts the leading dot-separated numeric components of a version (e.g. {@code "3.0.15"} from
+     * {@code "3.0.15.RELEASE"}) as a string {@link GradleVersion} can parse, dropping any trailing
```
mipo256: That is just wrong. We should not 

1. Try to parse the version right in here
2. And try to parse it in a way that conveys to `GradleVersion`

We have the `VersionTrimmer` that we used when we need to parse the software SemVer. I think the time has come for us to introduce the `SematicVersion` class.

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
mipo256, line 1: So, a couple of words for this class overall.

Do not write javadoc for private methods where it is obvious what they do. Simply don't. Because Javadoc can become stale, and it may lie about what the method does. When the behavior of the method is obvious - simply do not put javadoc. In case of private methods, write it only in cases **where you think it is worth explianing something that is not directly obvious**.

For public methods, though, the ones that represent the contract - the javadoc is pretty much mandatory or very desirable.

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/TestDependencyContributor.java
```diff
@@ -0,0 +1,152 @@
...
+    static final String PROFILER_NAME = "spring-test-profiler";
+
+    static final String THYMELEAF_DEPENDENCY = "org.thymeleaf:thymeleaf:3.1.3.RELEASE";
+
+    static final String THYMELEAF_GROUP = "org.thymeleaf";
+
+    static final String THYMELEAF_NAME = "thymeleaf";
+
+    /** Minimum Thymeleaf version the profiler's HTML report renders on. */
+    private static final GradleVersion MIN_THYMELEAF_VERSION = GradleVersion.version("3.1.3");
```
mipo256: Mate, let's not use the `GradleVersion` to represent the version of thymeleaf, this is just a non starter. The thymeleaf version

1. Does NOT follow the same format as thymeleaf and probably never would
2. It is very obscure for the reader on why are we doing that.

It is becoming obvious that we need a seprate SemVer class or something

mipo256 (changes requested):
The overall direction is good. The size of the PR is generally fine, good've been a bit less, but that is fine.

Still, there are some issues left, and I think we need to first think about incorporating the `SemVer` class into Axelix project. I will create a separte issue for that.

mipo256:
SemVer ticket https://github.com/axelixlabs/axelix/issues/1261

plugins/axelix-gradle-plugin/src/main/java/com/axelixlabs/gradle/plugin/AxelixGradlePlugin.java
```diff
@@ -39,6 +38,12 @@ public void apply(final Project project) {
...
 
     private void configure(Project project) {
-        // Configuration is added incrementally in subsequent changes.
+        InternalConfiguration configuration = new InternalConfiguration(project);
+
+        TestClasspathDependencyApplier testClasspathDependencyApplier = configuration.getTestClasspathDependencyApplier();
+        project.getConfigurations().getByName("testRuntimeClasspath").withDependencies(testClasspathDependencyApplier);
+
+        CompileClasspathDependencyApplier compileClasspathDependencyApplier = configuration.getCompileClasspathDependencyApplier();
+        project.getConfigurations().getByName("compileClasspath").withDependencies(compileClasspathDependencyApplier);
```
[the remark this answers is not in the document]
chrshnv: thymeleaf is usually declared in compileClasspath

Issue #1245, opened 2026-06-16
[Feature] Implement a new Dashboard design


mipo256:
On the dashboard page, we need to introduce the sidebar with the following entries:

```
- Overview
- Technologies
   - Java
   - Persistence
   - Spring Framework
```

The sider categories should be expanded by default (i.e. "Technologies" should be expanded by default).

On the "Overview" page we should have the same content as we now have on the Dashboard.

On the "Java" page we should have the following charts:

- "Project Leyden Adoption": The Donut Chart. 
- "Garbage Collector Distribution": The Donut Chart
- "Garbage Collector Logging": The Gauge Chart: 2 categories

On the "Persistence" page we should have the following charts:

- "N + 1": Treemap. 
- "In-Memory Pagination": Treemap.

On the "Spring Framework" page we should have the following charts:

- "OSIV": The Gauge Chart: 2 categories

Issue #1222, opened 2026-06-14
[Feature] Create a maven plugin for Axelix


mipo256:
I guess the task will be more or less the same as #1218. This is also supposed to be a Gradle project, included into the monorepo gradle build of Axelix. Semantics are similar to those explaied in #1218, the exception would be the GAV. They should be:

GroupID: `com.axelixlabs`
ArtifactId: `axelix-maven-plugin`

chrshnv:
assign me

PR #1253, opened 2026-06-16
GH-1252 Added applicationId for grouping instances

### LINKED ISSUE #1252: [Feature]  Add applicationId to Instance and self-registration metadata for grouping sibling instances

sergeycherkasovv:
## Release Notes

* **New Features**
  * Instances can now be grouped by application identifier for better organization and management.
  * Build metadata is automatically captured and used to identify associated instances.
  * Instance API responses now include application association details.

### DIFF
--- a/common/api/src/main/java/com/axelixlabs/axelix/common/api/registration/SelfRegistrationMetadata.java
+++ b/common/api/src/main/java/com/axelixlabs/axelix/common/api/registration/SelfRegistrationMetadata.java
@@ -21,6 +21,7 @@ import java.util.Objects;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.jspecify.annotations.Nullable;
 
 /**
  * Represents extended service instance metadata supplied as part of the self-registration process.
@@ -33,6 +34,7 @@ public class SelfRegistrationMetadata {
 
     private final BasicDiscoveryMetadata basicDiscoveryMetadata;
     private final String instanceId;
+    private final @Nullable String applicationId;
     private final String instanceName;
     private final String instanceActuatorUrl;
     private final String deploymentAt;
@@ -42,6 +44,10 @@ public class SelfRegistrationMetadata {
      *
      * @param basicDiscoveryMetadata    the basic metadata of a service instance
      * @param instanceId                unique identifier (uid) of the service
+     * @param applicationId             identifier of the application this instance belongs to — a hash of the
+     *                                  service build coordinates (GAV). Instances sharing the same
+     *                                  {@code applicationId} are grouped as one application. May be {@code null}
+     *                                  when build information is unavailable.
      * @param instanceName              name of the service
      * @param instanceActuatorUrl       the URL of the service, including the postfix for the actuator path,
      *                                  e.g. {@code https://my-app:6061/actuator}.
@@ -51,11 +57,13 @@ public class SelfRegistrationMetadata {
     public SelfRegistrationMetadata(
             @JsonProperty("basicDiscoveryMetadata") BasicDiscoveryMetadata basicDiscoveryMetadata,
             @JsonProperty("instanceId") String instanceId,
+            @JsonProperty("applicationId") @Nullable String applicationId,
             @JsonProperty("instanceName") String instanceName,
             @JsonProperty("instanceActuatorUrl") String instanceActuatorUrl,
             @JsonProperty("deploymentAt") String deploymentAt) {
         this.basicDiscoveryMetadata = basicDiscoveryMetadata;
         this.instanceId = instanceId;
+        this.applicationId = applicationId;
         this.instanceName = instanceName;
         this.instanceActuatorUrl = instanceActuatorUrl;
         this.deploymentAt = deploymentAt;
@@ -67,6 +75,10 @@ public class SelfRegistrationMetadata {
 
     public String getInstanceId() {
         return instanceId;
+    }
+
+    public @Nullable String getApplicationId() {
+        return applicationId;
     }
 
     public String getInstanceName() {
@@ -89,6 +101,7 @@ public class SelfRegistrationMetadata {
         SelfRegistrationMetadata that = (SelfRegistrationMetadata) o;
         return Objects.equals(basicDiscoveryMetadata, that.basicDiscoveryMetadata)
                 && Objects.equals(instanceId, that.instanceId)
+                && Objects.equals(applicationId, that.applicationId)
                 && Objects.equals(instanceName, that.instanceName)
                 && Objects.equals(instanceActuatorUrl, that.instanceActuatorUrl)
                 && Objects.equals(deploymentAt, that.deploymentAt);
@@ -96,7 +109,8 @@ public class SelfRegistrationMetadata {
 
     @Override
     public int hashCode() {
-        return Objects.hash(basicDiscoveryMetadata, instanceId, instanceName, instanceActuatorUrl, deploymentAt);
+        return Objects.hash(
+                basicDiscoveryMetadata, instanceId, applicationId, instanceName, instanceActuatorUrl, deploymentAt);
     }
 
     @Override
@@ -106,6 +120,9 @@ public class SelfRegistrationMetadata {
                 + basicDiscoveryMetadata
                 + ", instanceId='"
                 + instanceId
+                + '\''
+                + ", applicationId='"
+                + applicationId
                 + '\''
                 + ", instanceName='"
                 + instanceName
--- a/master/src/main/java/com/axelixlabs/axelix/master/api/external/response/InstancesGridResponse.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/api/external/response/InstancesGridResponse.java
@@ -25,18 +25,22 @@ import org.jspecify.annotations.Nullable;
  * The grid of {@link InstanceShortProfile instances}.
  *
  * @author Mikhail Polivakha
+ * @author Sergey Cherkasov
  */
 public record InstancesGridResponse(Collection<InstanceShortProfile> instances) {
 
     /**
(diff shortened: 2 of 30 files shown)


common/api/src/main/java/com/axelixlabs/axelix/common/api/registration/SelfRegistrationMetadata.java
```diff
@@ -51,11 +57,13 @@ public class SelfRegistrationMetadata {
     public SelfRegistrationMetadata(
             @JsonProperty("basicDiscoveryMetadata") BasicDiscoveryMetadata basicDiscoveryMetadata,
             @JsonProperty("instanceId") String instanceId,
+            @JsonProperty("applicationId") @Nullable String applicationId,
```
sergeycherkasovv, line 60: `applicationId` may be `null` because when generating `applicationId` as `hash(${group:artifactId:version})` on the starter side during `self-registration`, we depend on the [build plugin](https://docs.spring.io/spring-boot/how-to/build.html#howto.build.generate-info). If it is not configured, we simply cannot obtain the required information. I looked into ways to remove this dependency on the plugin, but based on what I found, there is no straightforward solution.

master/src/main/java/com/axelixlabs/axelix/master/api/external/response/InstancesGridResponse.java
```diff
@@ -25,18 +25,22 @@
...
      *
+     * @param applicationId identifier of the application this instance belongs to; instances of the same
+     *                      application share it. May be {@code null}.
      * @param deployedFor a String representation for how long the service has been already deployed for.
      *
      * @author Mikhail Polivakha
      */
     public record InstanceShortProfile(
             String instanceId,
+            @Nullable String applicationId,
```
sergeycherkasovv, line 43: the same applies to the dependency on the build plugin)

master/src/main/java/com/axelixlabs/axelix/master/domain/Instance.java
```diff
@@ -48,6 +51,7 @@
 @Table("instances")
 public record Instance(
         @Id InstanceId id,
+        @Nullable String applicationId,
```
sergeycherkasovv, line 54: the same applies to the dependency on the build plugin))

sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultBuildInfoProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ * {@code META-INF/build-info.properties}), every getter returns an empty string.
+ *
+ * @author Sergey Cherkasov
+ */
+public class DefaultBuildInfoProvider implements BuildInfoProvider {
+    private final @Nullable BuildProperties buildProperties;
+
+    public DefaultBuildInfoProvider(@Nullable BuildProperties buildProperties) {
+        this.buildProperties = buildProperties;
+    }
```
sergeycherkasovv, line 39: We also depend on the `build plugin` in [DefaultServiceDetailsAssembler](https://github.com/axelixlabs/axelix/blob/master/sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/details/DefaultServiceDetailsAssembler.java#L47), where `BuildDetails` are created. So at this point, we rely on the plugin in two places. If we go with the `BuildInfoProvider` approach I suggested for retrieving build information, we could move `DefaultServiceDetailsAssemble` into `starter-domain`.

I also noticed that our starter setup [docs](https://axelix.io/docs/setting-up-spring-boot-service/configuring-axelix-starter/) do not mention that users need to configure the `git` and `build` plugins for the application to work properly. We should probably add that.

WYDT, @mipo256?

sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultBuildInfoProviderTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration test for {@link DefaultBuildInfoProvider}.
+ *
+ * @author Sergey Cherkasov
+ */
+@SpringBootTest
+@Import({DefaultBuildInfoProvider.class, ProjectInfoAutoConfiguration.class})
+@TestPropertySource(properties = "spring.info.build.location=classpath:other/build-info.properties")
```
sergeycherkasovv, line 37: I decided not to add this property to `application.yaml`, as is done for `git` testing, because it is used only in a single test class))

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultSelfRegistrationMetadataAssembler.java
```diff
@@ -53,8 +58,24 @@ public SelfRegistrationMetadata assemble() {
...
+        String version = buildInfoProvider.getVersion();
+
+        if (artifact.isEmpty() && group.isEmpty() && version.isEmpty()) {
+            return null;
+        }
+
+        String applicationId = artifact + ":" + group + ":" + version;
+
+        return String.valueOf(UUID.nameUUIDFromBytes(applicationId.getBytes(StandardCharsets.UTF_8)));
+    }
```
sergeycherkasovv, line 80: I think if the user doesn’t have the `build` plugin configured, and `artifact`, `group`, and `version` are all **empty**, then `applicationId` should be set to `null`. That way, we avoid accidentally assigning the `instance` to a group. If we hash **empty** `artifact`, `group`, and `version` values, multiple `instances` could end up with the same `applicationId` and be incorrectly merged into one group, which would mislead users.

I think we could change the `applicationId` generation logic to something like this:
1. If `artifact`, `group`, and `version` are **present**, generate `hash(${group:artifactId:version})`
2. If `artifact`, `group`, and `version` are **empty**, use `spring.application.name` or `axelix.sbs.discovery.instance-name` and generate `hash(${name})` from the instance name.
3. If nothing is known, return `null`.

WYDT, @mipo256?

sbs/axelix-spring-boot-2-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultBuildInfoProviderTest.java
```diff
@@ -0,0 +1,61 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration test for {@link DefaultBuildInfoProvider}.
+ *
+ * @author Sergey Cherkasov
+ */
+@SpringBootTest
+@Import({DefaultBuildInfoProvider.class, ProjectInfoAutoConfiguration.class})
+@TestPropertySource(properties = "spring.info.build.location=classpath:other/build-info.properties")
```
sergeycherkasovv, line 37: I decided not to add this property to `application.yaml`, as is done for `git` testing, because it is used only in a single test class))

sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultBuildInfoProvider.java
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * @author Sergey Cherkasov
+ */
+public class DefaultBuildInfoProvider implements BuildInfoProvider {
+    private final @Nullable BuildProperties buildProperties;
+
+    public DefaultBuildInfoProvider(@Nullable BuildProperties buildProperties) {
+        this.buildProperties = buildProperties;
+    }
+
```
sergeycherkasovv, line 40: The same applies here, if we keep the `BuildInfoProvide`r interface-based approach, we can move [DefaultServiceDetailsAssembler](https://github.com/axelixlabs/axelix/blob/master/sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/details/DefaultServiceDetailsAssembler.java#L47) into `starter-domain`)

common/api/src/main/java/com/axelixlabs/axelix/common/api/registration/SelfRegistrationMetadata.java
mipo256: I understand the concern. 

However, I think that if we cannot report the `applicationId`, we should not register the instance at all, I think. We're going to introduce our own plugins system very soon (see #1222 and #1218), and thus Axelix would rely on infromation that is provided by our build plugin, not by `git-commit-id` or `spring-boot` plugins for maven/gradle.

master/src/main/resources/db/changelog/mysql/changelog/1.0.0-create-instances-table.xml
```diff
@@ -17,6 +17,9 @@
       <column name="instance_id" type="VARCHAR(768)">
         <constraints nullable="false" primaryKey="true" primaryKeyName="pk_instances"/>
       </column>
+      <column name="application_id" type="VARCHAR(1000)">
```
mipo256, line 20: We no longer can do that, we're going to break existing deployments if we do that.

Create a new migration. Ditto for evey other database.

mipo256 (changes requested):
I think this PR needs to wait until we will ship the plugins system. I understand the motivation behind nullable application id, but I think that this is going to introduce unnecessary duality and complication. 

Ideally, for the crucial things like `applicationId` we should never such duality - can/cannot it be null. Apart from that, as the dashboard story continues (see #1245 ), I think, we are going to revisit the `applicaitonId` design.

mipo256:
We have chosen another approach, that change is no longer needed

PR #1256, opened 2026-06-17
test(sbs3): run AxelixConfigurationPropertiesEndpointTest in shared context

chrshnv:
## Summary

Folds `AxelixConfigurationPropertiesEndpointTest` into the shared, cached configprops Spring context introduced in #1233 (`AbstractConfigPropsSharedContextTest`). The endpoint test previously declared its own standalone `@SpringBootTest(webEnvironment = RANDOM_PORT)` and thus loaded a separate Spring context. Now all three configprops integration tests resolve to a single cached context.

## Changes

- **`AbstractConfigPropsSharedContextTest`** — switch to `RANDOM_PORT`, import `JwtAuthTestConfiguration`, and add the `forSanitization`/`FOR_SANITIZATION` tags and `requests[1]` properties so the merged context configuration is the superset all three integration tests need.
- **`ConfigPropsTestSupportConfiguration`** — add a `DefaultConfigurationPropertiesService` that sanitizes the `forSanitization` tags and register the `AxelixConfigurationPropertiesEndpoint` wired to it; drop the duplicate `securityContextExecutor` bean (now provided by the JWT config's `@ConditionalOnMissingBean`).
- **`AxelixConfigurationPropertiesEndpointTest`** — extend the shared base; remove its own `@SpringBootTest`/`@TestPropertySource`/`@EnableConfigurationProperties`/`@Import`, inner `@TestConfiguration` and nested record. Test logic unchanged.
- **`DefaultConfigurationPropertiesConverterTest`** — expand assertions for the larger shared property set (`inputs` 20 → 34).

## Verification

- `./gradlew :sbs:axelix-spring-boot-3-starter:test` — green.
- spring-test-profiler now reports all three configprops integration tests on a **single** cached context (1 cache miss, 96.7% hit rate); previously the endpoint test loaded its own.
- Spotless + PMD + NullAway pass.

## Summary by CodeRabbit

* **Tests**
  * Consolidated configuration-properties endpoint tests under a shared Spring Boot test base (using a random port) to reduce duplication.
  * Expanded sanitization coverage by adding additional tag key variants and adjusting expected tag/value assertions by role.
  * Updated and broadened converter/service test expectations, including additional HTTP client request properties and expanded input/property sets.

* **Chores**
  * Simplified and aligned test wiring by removing unused security context setup and explicitly targeting the correct sanitizing configuration properties service.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/ConfigPropsTestSupportConfiguration.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/ConfigPropsTestSupportConfiguration.java
@@ -20,6 +20,7 @@ package com.axelixlabs.axelix.sbs.spring.core.configprops;
 import java.util.List;
 import java.util.Map;
 
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.test.context.TestConfiguration;
@@ -28,7 +29,6 @@ import org.springframework.context.annotation.Bean;
 
 import com.axelixlabs.axelix.common.auth.core.SecurityContextExecutor;
 import com.axelixlabs.axelix.sbs.spring.core.auth.RequiredAuthorityCheckService;
-import com.axelixlabs.axelix.sbs.spring.core.auth.ThreadLocalSecurityContextExecutor;
 import com.axelixlabs.axelix.sbs.spring.core.config.EndpointsConfigurationProperties;
 import com.axelixlabs.axelix.sbs.spring.core.configprops.ConfigPropsTestSupportConfiguration.SharedAxelixConfigurationProperties;
 import com.axelixlabs.axelix.sbs.spring.core.env.DefaultPropertyNameNormalizer;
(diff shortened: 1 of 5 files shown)


sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/configprops/ConfigPropsTestSupportConfiguration.java
```diff
@@ -114,6 +112,29 @@ public DefaultConfigurationPropertiesService explicitlySanitizedConfigurationPro
                 requiredAuthorityCheckService);
     }
 
+    @Bean(name = FOR_SANITIZATION_TAGS_CONFIGURATION_PROPERTIES_SERVICE)
```
mipo256: There is no need in this bean acutally. There is a bean that sintizes these properties already.

```
axelix.prop.test.tags.environment=test
axelix.prop.test.tags.version=1.0.0
```

creating yet another bean that sanitizes another set of properties is possible, but that does not make much sense - it creates more code and confused the reader - why would we have two different beans that just sanitze different set of properties? 

Let's not add yet another one `DefaultConfigurationPropertiesService` and stick with the existing one. That probably will imply changing `AxelixConfigurationPropertiesEndpointTest` and `DefaultConfigurationPropertiesConverterTest` a bit

mipo256 (changes requested):
I have one thing to say, apart from that, the change is good.

PR #1259, opened 2026-06-17
GH-1258 Added support for multiple instances for a single logger when updating levels

### LINKED ISSUE #1258: [Feature] Support for multiple instances when managing level updates for a single logger in the main API

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Added bulk logger management support across multiple instances, including changing logger levels, changing logger groups, and resetting loggers.
  * Introduced clearer API responses for partial updates.

* **Bug Fixes**
  * Bulk requests now ignore blank or duplicate instance IDs and validate required fields more strictly.
  * Partial failures now return a client-facing error instead of failing silently.

* **Tests**
  * Expanded coverage for multi-instance success, partial failure, invalid input, and endpoint invocation behavior.

### DIFF
--- a/master/src/main/java/com/axelixlabs/axelix/master/api/external/ApiPaths.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/api/external/ApiPaths.java
@@ -151,6 +151,11 @@ public final class ApiPaths {
         public static final String LOGGER_NAME = "/loggers/{instanceId}/logger/{loggerName}";
 
         /**
+         * Endpoint to change a specific logger by name across instances.
+         */
+        public static final String LOGGER_BULK_CHANGE = "/loggers/logger";
+
+        /**
          * Endpoint to retrieve a specific logger group by name from an instance.
          */
         public static final String GROUP_NAME = "/loggers/{instanceId}/group/{groupName}";
--- a/master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
@@ -20,16 +20,13 @@ package com.axelixlabs.axelix.master.api.external.endpoint;
 import java.util.List;
 import java.util.Map;
 
-import com.axelixlabs.axelix.master.api.error.handle.ApiErrorCodes;
-import com.axelixlabs.axelix.master.service.transport.BadRequestException;
-import com.axelixlabs.axelix.master.service.transport.PartiallyUpdatedException;
 import io.swagger.v3.oas.annotations.Parameter;
 import io.swagger.v3.oas.annotations.media.Content;
 import io.swagger.v3.oas.annotations.media.Schema;
 import io.swagger.v3.oas.annotations.responses.ApiResponse;
 import io.swagger.v3.oas.annotations.tags.Tag;
-
 import org.jspecify.annotations.Nullable;
+
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
@@ -47,6 +44,7 @@ import com.axelixlabs.axelix.common.domain.http.DefaultHttpPayload;
 import com.axelixlabs.axelix.common.domain.http.HttpPayload;
 import com.axelixlabs.axelix.common.domain.http.NoHttpPayload;
 import com.axelixlabs.axelix.master.api.error.SimpleApiError;
+import com.axelixlabs.axelix.master.api.error.handle.ApiErrorCodes;
 import com.axelixlabs.axelix.master.api.external.ApiPaths;
 import com.axelixlabs.axelix.master.api.external.ExternalApiRestController;
 import com.axelixlabs.axelix.master.api.external.request.loggers.LogLevelLoggerBulkChangeRequest;
@@ -56,7 +54,9 @@ import com.axelixlabs.axelix.master.api.external.swagger.DefaultApiResponse;
 import com.axelixlabs.axelix.master.api.external.swagger.InstanceIdParameter;
 import com.axelixlabs.axelix.master.domain.InstanceId;
 import com.axelixlabs.axelix.master.service.serde.JacksonMessageSerializationStrategy;
+import com.axelixlabs.axelix.master.service.transport.BadRequestException;
 import com.axelixlabs.axelix.master.service.transport.EndpointInvoker;
+import com.axelixlabs.axelix.master.service.transport.PartiallyUpdatedException;
 
 /**
  * The API for managing loggers.
@@ -136,7 +136,7 @@ public class LoggersApi {
             summary = "Change the logging level for a given logger by its name across instances.",
             description =
                     "Suggested logging levels that the user can select to configure the logger: OFF, FATAL, ERROR, WARN, INFO, DEBUG, TRACE")
-    @ApiResponse(description = "OK", responseCode = "200")
+    @ApiResponse(description = "No Content", responseCode = "204")
     @PostMapping(path = ApiPaths.LoggersApi.LOGGER_BULK_CHANGE)
     public ResponseEntity<?> setLoggingLevelByLoggerName(@RequestBody LogLevelLoggerBulkChangeRequest request) {
 
@@ -146,15 +146,16 @@ public class LoggersApi {
 
         HttpPayload payload = HttpPayload.json(
                 Map.of("name", request.loggerName()),
-                jacksonMessageSerializationStrategy.serialize(new LogLevelChangeRequest(request.configuredLevel(), request.ttlMinutes())));
+                jacksonMessageSerializationStrategy.serialize(
+                        new LogLevelChangeRequest(request.configuredLevel(), request.ttlSeconds())));
 
         try {
             endpointInvoker.invokeForInstances(request.instanceIds(), ActuatorEndpoints.SET_ONE_LOGGER, payload);
             return ResponseEntity.noContent().build();
         } catch (PartiallyUpdatedException e) {
             return ResponseEntity.badRequest()
-                .body(new SimpleApiError(
-                    ApiErrorCodes.PARTIALLY_UPDATED.getErrorCode(), HttpStatus.BAD_REQUEST.value()));
+                    .body(new SimpleApiError(
+                            ApiErrorCodes.PARTIALLY_UPDATED.getErrorCode(), HttpStatus.BAD_REQUEST.value()));
         } catch (BadRequestException e) {
             return ResponseEntity.badRequest().build();
         }
@@ -194,7 +195,10 @@ public class LoggersApi {
         return ResponseEntity.noContent().build();
     }
 
-    private boolean isInvalid(@Nullable List<String> instanceIds, @Nullable String name, @Nullable String configuredLevel) {
-        return CollectionUtils.isEmpty(instanceIds) || !StringUtils.hasText(name) || !StringUtils.hasText(configuredLevel);
+    private boolean isInvalid(
+            @Nullable List<String> instanceIds, @Nullable String name, @Nullable String configuredLevel) {
+        return CollectionUtils.isEmpty(instanceIds)
+                || !StringUtils.hasText(name)
+                || !StringUtils.hasText(configuredLevel);
     }
 }
--- a/master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
@@ -54,7 +54,6 @@ import com.axelixlabs.axelix.master.api.external.swagger.DefaultApiResponse;
 import com.axelixlabs.axelix.master.api.external.swagger.InstanceIdParameter;
 import com.axelixlabs.axelix.master.domain.InstanceId;
 import com.axelixlabs.axelix.master.service.serde.JacksonMessageSerializationStrategy;
-import com.axelixlabs.axelix.master.service.transport.BadRequestException;
 import com.axelixlabs.axelix.master.service.transport.EndpointInvoker;
 import com.axelixlabs.axelix.master.service.transport.PartiallyUpdatedException;
 
@@ -150,14 +149,12 @@ public class LoggersApi {
                         new LogLevelChangeRequest(request.configuredLevel(), request.ttlSeconds())));
 
         try {
-            endpointInvoker.invokeForInstances(request.instanceIds(), ActuatorEndpoints.SET_ONE_LOGGER, payload);
+            endpointInvoker.invokeNoValueForInstances(request.instanceIds(), ActuatorEndpoints.SET_ONE_LOGGER, payload);
             return ResponseEntity.noContent().build();
         } catch (PartiallyUpdatedException e) {
             return ResponseEntity.badRequest()
                     .body(new SimpleApiError(
                             ApiErrorCodes.PARTIALLY_UPDATED.getErrorCode(), HttpStatus.BAD_REQUEST.value()));
-        } catch (BadRequestException e) {
-            return ResponseEntity.badRequest().build();
         }
     }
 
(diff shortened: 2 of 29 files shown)


master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
```diff
@@ -123,54 +131,127 @@ public ResponseEntity<byte[]> getLoggerByName(
...
-                InstanceId.of(instanceId),
-                ActuatorEndpoints.RESET_FOR_LOGGER,
-                new DefaultHttpPayload(Map.of("name", loggerName)));
+        for (String instanceId : instanceIds) {
+            try {
+                endpointInvoker.invokeNoValue(InstanceId.of(instanceId), endpoint, payload);
+            } catch (RuntimeException e) {
+                failures++;
+            }
+        }
```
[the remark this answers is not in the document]
sergeycherkasovv: done)

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApiManagementLoggingLevelTest.java
```diff
@@ -56,14 +63,16 @@
...
-    private static final String LOG_LEVEL_LOGGER_AUTH_JSON = "{\"configuredLevel\":\"DEBUG\"}";
-
     private static final String activeInstanceId = UUID.randomUUID().toString();
+    private static final String siblingInstanceId = UUID.randomUUID().toString();
+    private static final String failingInstanceId = UUID.randomUUID().toString();
 
     private static MockWebServer mockWebServer;
 
+    private Set<String> invokedPaths;
+
```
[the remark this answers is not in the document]
sergeycherkasovv: done)

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApiManagementLoggingLevelTest.java
```diff
@@ -83,227 +90,241 @@ static void shutdownServer() throws IOException {
...
-            method = HttpMethod.POST,
-            path = "/api/external/loggers/00000000-0000-0000-0000-000000000001/logger/clear.logger.name/clear")
-    void negativeAuthTestsOnClearLoggingLevelByLoggerName() {}
+    @ProtectedEndpointTests(method = HttpMethod.POST, path = "/api/external/loggers/logger/reset", jsonBody = """
+                    {
+                      "instanceIds": ["00000000-0000-0000-0000-000000000001"],
+                      "loggerName": "reset.logger.name"
+                    }
+                    """)
+    void negativeAuthTestsOnResetLoggingLevelByLoggerName() {}
```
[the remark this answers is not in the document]
sergeycherkasovv: I think we don’t need this :)

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApiManagementLoggingLevelTest.java
```diff
@@ -83,227 +90,241 @@ static void shutdownServer() throws IOException {
...
+        ResponseEntity<String> response =
+                restTemplate.asViewer().postForEntity("/api/external/loggers/logger", requestBody, String.class);
+
+        // then.
+        assertThat(response.getStatusCode()).isEqualTo(HttpStatus.BAD_REQUEST);
+        assertThat(invokedPaths)
+                .containsExactlyInAnyOrder(
+                        "/" + activeInstanceId + "/axelix-loggers/logger/logger.name/change-level",
+                        "/" + failingInstanceId + "/axelix-loggers/logger/logger.name/change-level");
     }
```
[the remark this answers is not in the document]
sergeycherkasovv: done)

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApiManagementLoggingLevelTest.java
```diff
@@ -56,14 +61,16 @@
...
     private static final String activeInstanceId = UUID.randomUUID().toString();
+    private static final String siblingInstanceId = UUID.randomUUID().toString();
+    private static final String failingInstanceId = UUID.randomUUID().toString();
 
     private static MockWebServer mockWebServer;
 
+    private List<String> invokedPaths;
+
+    private int failingLoggerUpdateStatusCode;
+
```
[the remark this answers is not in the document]
sergeycherkasovv: done)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
```diff
@@ -123,54 +134,155 @@ public ResponseEntity<byte[]> getLoggerByName(
...
+                                            schema = @Schema(implementation = LogLevelLoggerBulkChangeRequest.class)))
+                    @RequestBody
+                    LogLevelLoggerBulkChangeRequest request) {
+
+        if (request == null
+                || isNullOrEmpty(request.instanceIds())
+                || isNullOrBlank(request.loggerName())
+                || isNullOrBlank(request.configuredLevel())) {
+            return ResponseEntity.badRequest().build();
+        }
```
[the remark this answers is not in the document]
sergeycherkasovv: Skipping. This is intentional for the bulk logger fan-out: the **API** rejects a null or empty instanceIds collection, but ignores `null/blank` entries inside a non-empty collection before deduplicating and invoking instances. This prevents invalid entries from causing duplicate or malformed downstream calls while still allowing the valid requested instances to be updated.

master/src/main/java/com/axelixlabs/axelix/master/service/transport/DefaultEndpointInvoker.java
```diff
@@ -71,6 +73,36 @@ public void invokeNoValue(InstanceId instanceId, ActuatorEndpoint endpoint, Http
...
+            }
+        }
+
+        int instanceIdsSize = uniqueInstanceIds.size();
+
+        if (failures > 0 && failures < instanceIdsSize) {
+            throw new PartiallyUpdatedException("Some instances failed to update");
+        } else if (failures == instanceIdsSize) {
+            throw new BadRequestException("All instances failed to update");
+        }
```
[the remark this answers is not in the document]
sergeycherkasovv: Skipped. Current behavior is intentional: invokeForInstances treats an all-failed bulk operation as a bad bulk request regardless of the per-instance failure mix. A single bulk request can contain failures of different types at the same time (EndpointInvocationException, BadRequestException, InstanceNotFoundException), so preserving one arbitrary exception type would be misleading.

master/src/test/java/com/axelixlabs/axelix/master/service/transport/DefaultEndpointInvokerTest.java
```diff
@@ -237,6 +240,84 @@ void invokeNoValue_shouldReturnBadRequestException_OnNotValidRequest() {
...
+
+        // when.
+        ThrowableAssert.ThrowingCallable callable = () -> subject.invokeForInstances(
+                List.of("first-instance"), METHOD_INVOKE, NoHttpPayload.INSTANCE);
+
+        // then.
+        assertThatThrownBy(callable).isInstanceOf(EndpointInvocationException.class);
+        assertThat(prober.invokedInstanceIds()).isEmpty();
+    }
+
```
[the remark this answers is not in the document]
sergeycherkasovv: Skipped. Not this time)

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
```diff
@@ -123,21 +133,32 @@ public ResponseEntity<byte[]> getLoggerByName(
...
-
-        HttpPayload payload =
-                HttpPayload.json(Map.of("name", loggerName), jacksonMessageSerializationStrategy.serialize(request));
-        endpointInvoker.invokeNoValue(InstanceId.of(instanceId), ActuatorEndpoints.SET_ONE_LOGGER, payload);
+    @PostMapping(path = ApiPaths.LoggersApi.LOGGER_BULK_CHANGE)
+    public ResponseEntity<?> setLoggingLevelByLoggerName(@RequestBody LogLevelLoggerBulkChangeRequest request) {
+
+        if (request == null || isInvalid(request.instanceIds(), request.loggerName(), request.configuredLevel())) {
+            return ResponseEntity.badRequest().build();
+        }
```
sergeycherkasovv, line 144: I fully understand that the only source of this data will be the `UI`, but I think we should also consider that our service might be used via tools like `Postman`. Maybe I’m overthinking it

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
```diff
@@ -173,4 +194,11 @@ public ResponseEntity<Void> resetLoggingLevelByLoggerName(
 
         return ResponseEntity.noContent().build();
     }
+
+    private boolean isInvalid(
+            @Nullable List<String> instanceIds, @Nullable String name, @Nullable String configuredLevel) {
+        return CollectionUtils.isEmpty(instanceIds)
+                || !StringUtils.hasText(name)
+                || !StringUtils.hasText(configuredLevel);
+    }
```
sergeycherkasovv, line 200: This method will also be used in the `setLoggingLevelByGroupName` endpoint

master/src/main/java/com/axelixlabs/axelix/master/api/external/ApiPaths.java
```diff
@@ -140,6 +140,11 @@ public static final class LoggersApi {
          */
         public static final String LOGGER_NAME = "/loggers/{instanceId}/logger/{loggerName}";
 
+        /**
+         * Endpoint to change a specific logger by name across instances.
+         */
+        public static final String LOGGER_BULK_CHANGE = "/loggers/logger";
+
```
sergeycherkasovv, line 157: In this case, we specify a list of `instanceIds` in the request body `LogLevelLoggerBulkChangeRequest`

master/src/main/java/com/axelixlabs/axelix/master/service/transport/DefaultEndpointInvoker.java
```diff
@@ -71,6 +73,34 @@ public void invokeNoValue(InstanceId instanceId, ActuatorEndpoint endpoint, Http
         getEndpointProber(endpoint).invoke(instanceId, httpPayload);
     }
 
+    @Override
+    public void invokeForInstances(List<String> instanceIds, ActuatorEndpoint endpoint, HttpPayload payload)
+            throws PartiallyUpdatedException, BadRequestException {
+
+        List<String> uniqueInstanceIds =
+                instanceIds.stream().filter(StringUtils::hasText).distinct().toList();
```
sergeycherkasovv, line 81: It seems to me that we have two possible approaches when the list of instanceIds contains invalid entries:

1. Filter them out and proceed using only valid instanceIds  
2. If at least one instanceId is invalid, throw an exception . The second option feels more correct, as it provides clear feedback (for example, when using MCP or Postman).

However, I chose the first approach because it’s less intrusive. If a user makes a mistake in MCP or Postman, we won’t fail the request, since it’s not a critical error and doesn’t necessarily require their attention.

master/src/main/java/com/axelixlabs/axelix/master/service/transport/DefaultEndpointInvoker.java
```diff
@@ -71,6 +73,34 @@ public void invokeNoValue(InstanceId instanceId, ActuatorEndpoint endpoint, Http
         getEndpointProber(endpoint).invoke(instanceId, httpPayload);
     }
 
+    @Override
+    public void invokeForInstances(List<String> instanceIds, ActuatorEndpoint endpoint, HttpPayload payload)
```
sergeycherkasovv: After discussing with @NikitaKirilloff , we decided that `invokeForInstances()` should be placed in `EndpointInvoker`, as it is expected to be used in places beyond `LoggersApi`

master/src/main/java/com/axelixlabs/axelix/master/service/transport/PartiallyUpdatedException.java
```diff
@@ -0,0 +1,30 @@
...
+ * Indicates that a bulk endpoint invocation completed only for a subset of requested instances.
+ *
+ * @author Sergey Cherkasov
+ */
+public class PartiallyUpdatedException extends RuntimeException {
+
+    public PartiallyUpdatedException(String message) {
+        super(message);
+    }
+}
```
sergeycherkasovv, line 30: It’s possible that in the future we’ll be able to report exactly which instances the operation failed for))

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
```diff
@@ -123,21 +133,32 @@ public ResponseEntity<byte[]> getLoggerByName(
...
-    public void setLoggingLevelByLoggerName(
-            @PathVariable("instanceId") String instanceId,
-            @PathVariable("loggerName") String loggerName,
-            @RequestBody LogLevelChangeRequest request) {
-
-        HttpPayload payload =
-                HttpPayload.json(Map.of("name", loggerName), jacksonMessageSerializationStrategy.serialize(request));
-        endpointInvoker.invokeNoValue(InstanceId.of(instanceId), ActuatorEndpoints.SET_ONE_LOGGER, payload);
+    @PostMapping(path = ApiPaths.LoggersApi.LOGGER_BULK_CHANGE)
+    public ResponseEntity<?> setLoggingLevelByLoggerName(@RequestBody LogLevelLoggerBulkChangeRequest request) {
```
sergeycherkasovv, line 140: We also need to update how we interact with loggers in `LoggersMcpServerTools`, and it seems to me that the most effective approach would be to create `Skills` for updating and creating `MCP tools`

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApiManagementLoggingLevelTest.java
```diff
@@ -83,31 +93,46 @@ static void shutdownServer() throws IOException {
...
         mockWebServer.setDispatcher(new Dispatcher() {
             @Override
             public @NotNull MockResponse dispatch(@NotNull RecordedRequest request) {
                 String path = request.getPath();
                 assert path != null;
+                invokedPaths.add(path);
 
+                if (path.equals("/" + failingInstanceId + "/axelix-loggers/logger/logger.name/change-level")) {
+                    return new MockResponse().setResponseCode(failingLoggerUpdateStatusCode.get());
+                }
```
sergeycherkasovv: `failingLoggerUpdateStatusCode` lets the same mock endpoint simulate both downstream failure classes used by `invokeForInstances`: `400` from the managed app is handled as `BadRequestException`, while `500` is handled as `EndpointInvocationException`.

For example, `@ValueSource(ints = {400, 500})` runs the same test twice, and each run configures the mock response through `failingLoggerUpdateStatusCode.set(failureStatusCode)`. The dispatcher then reads that value via `failingLoggerUpdateStatusCode.get()`, so the first run exercises the downstream `400` path and the second run exercises the downstream `500` path.

This verifies that bulk logger update still invokes all requested instances and returns `PARTIALLY_UPDATED` for both cases. If the dispatcher always returned a hardcoded `return new MockResponse().setResponseCode(500)`, both parameterized runs would exercise the same server-error path and the downstream `400` scenario would no longer be covered.

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
mipo256: I am not exactly sure what skills we need to create here. I do not see the clear need for that.

However, I have some thoughts about creating the skill for writing MCP servers in general. But probably we're not going to need to write it ourselves - probably someone else has already created it and put it in Open Source.

master/src/main/java/com/axelixlabs/axelix/master/api/external/endpoint/LoggersApi.java
mipo256: I think that is alright.

mipo256 (commented):
I agree overall with the approach (placing multi-instance call inside `EndpointInvoker`). I however think we should do it in a slightly different way. 

I will make this PR to completion myself.

PR #1264, opened 2026-06-18
Share the beans cached context with the endpoint test

chrshnv:
## What

Folds `AxelixBeansEndpointTest` into the shared `AbstractBeansSharedContextTest` cached context, so the entire `beans` test package uses a **single** Spring `ApplicationContext` instead of two.

- The endpoint configuration (`CurrentConfiguration`, `AxelixPropTest`, the `RANDOM_PORT` web environment, the actuator endpoint imports and the test property source) is moved into the parent — which keeps owning the shared configuration directly and never references its subclasses.
- `AxelixBeansEndpointTest` becomes a thin subclass with only its test methods.
- The duplicate `conditionalBeanRefBuilder` / `beanMetaInfoExtractor` / `qualifiersPersistencePostProcessor` beans are removed from `DefaultBeanAnalyzerTestConfig`, since they would otherwise collide with `CurrentConfiguration`'s beans now that both live in the same context (bean overriding is disabled).

## Why

One cached context for the whole package → fewer context loads / faster test runs.

## Stacked on #1263

This PR builds on #1263 (configuration centralization) and should be merged **after** it. Until #1263 merges, the diff here also contains that commit — review the second commit, *"Share the beans cached context with the endpoint test"*, for the net change.

## Verification

- `spring-test-profiler` report shows all three beans test classes (`DefaultBeanMetaInfoExtractorTest`, `AxelixBeansEndpointTest`, `QualifiersPersistencePostProcessorTest`) sharing one context.
- Full `:sbs:axelix-spring-boot-3-starter:build` (tests + Spotless + PMD + NullAway) passes.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AbstractBeansSharedContextTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/beans/AbstractBeansSharedContextTest.java
@@ -22,6 +22,7 @@ import java.lang.annotation.ElementType;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
+import java.util.function.Supplier;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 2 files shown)


mipo256:
Please rebase @chrshnv

chrshnv:
@mipo256 done

mipo256 (approved):

PR #1265, opened 2026-06-18
GH-1261: Add SemanticVersion value object to common/utils

### LINKED ISSUE #1261: [Feature] Introduce the SemVer dedicated abstraction
After the review of https://github.com/axelixlabs/axelix/pull/1238 it became clear that we sometimes need to parse the semantic version. 

We already kind of have some api (see `VersionTrimmer`), but that is an ad-hoc api that was created in place just to quickly parse sem-vers.

I think, we need to replace the `VersionTrimmer` with the dedicated class, like `SemVer` that is gong to be able to parse sem vers (see [what they exactly are](https://semver.org/)), but apart from that it must be capable to strip things like `.RELEASE` or `.Final` postfixes that projects sometimes add.

chrshnv:
Part of #1261.

## What

Adds an immutable `SemanticVersion` value object to `common/utils`, so version parsing can be shared across `master` and the starter modules (today the only helper, `VersionTrimmer`, lives in `master` and can't strip qualifier postfixes).

## API

- `SemanticVersion.parse(String)` — strict, throws `IllegalArgumentException` on null/blank/non-numeric input.
- `SemanticVersion.tryParse(String)` — lenient, returns `Optional<SemanticVersion>`.
- Numeric accessors: `int major()`, `int minor()`, `int patch()`.
- Formatted views: `String majorMinor()`, `String versionNumber()` (numeric `major.minor.patch`), `@Nullable String qualifier()`.

## Parsing

At most three numeric segments are consumed (`major.minor.patch`); everything after `patch` is the qualifier, introduced by either a `.` or a `-` separator. So postfixes such as `.RELEASE`, `.Final` and `-SNAPSHOT` are recognised and stripped from the numeric accessors — e.g. `3.0.15.RELEASE` → major `3`, minor `0`, patch `15`, qualifier `RELEASE`. A trailing numeric segment beyond patch becomes the qualifier too (`1.2.3.4` → qualifier `4`).

Segmentation uses `indexOf('.')` + `substring` (mirroring the existing `VersionTrimmer` style).

## Notes

This PR adds the new abstraction only. Migrating the existing `VersionTrimmer` call sites onto `SemanticVersion` (and deleting `VersionTrimmer`) is the follow-up that closes #1261.

## Release Notes

* **New Features**
  * Added a new immutable `SemanticVersion` value object for parsing `major.minor.patch` versions with optional `.QUALIFIER` or `-QUALIFIER` suffixes.
  * Provides version component accessors and consistent ordering using numeric comparison, with deterministic handling of presence/absence of qualifiers.

* **Tests**
  * Added JUnit coverage for successful parsing, invalid input scenarios, component extraction (including qualifier formats), and `compareTo` ordering across numeric-only and qualified versions.

### DIFF
--- a/common/utils/src/main/java/com/axelixlabs/axelix/common/utils/SemanticVersion.java
+++ b/common/utils/src/main/java/com/axelixlabs/axelix/common/utils/SemanticVersion.java
@@ -19,6 +19,7 @@ package com.axelixlabs.axelix.common.utils;
 
 import java.util.Comparator;
 import java.util.Optional;
+import java.util.Set;
 
 import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
@@ -32,9 +33,13 @@ import org.jspecify.annotations.Nullable;
  * separator. So postfixes such as {@code .RELEASE}, {@code .Final} or {@code -SNAPSHOT} are
  * recognised and stripped from the numeric accessors (e.g. {@code 3.0.15.RELEASE} yields major
  * {@code 3}, minor {@code 0}, patch {@code 15} and qualifier {@code RELEASE}).
+ * <p>
+ * Note, that according to SemVer specification, the semantic version MUST have the major, minor and
+ * the patch. Missing any of those will result in parsing error.
  *
- * @author Nikita Kirillov
  * @author Artemiy Degtyarev
+ * @author Nikita Kirillov
+ * @author Mikhail Polivakha
  */
 public class SemanticVersion implements Comparable<SemanticVersion> {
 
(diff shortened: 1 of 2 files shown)


common/utils/src/main/java/com/axelixlabs/axelix/common/utils/SemanticVersion.java
```diff
@@ -0,0 +1,228 @@
...
+            .thenComparingInt(SemanticVersion::minor)
+            .thenComparingInt(SemanticVersion::patch)
+            // A qualifier (e.g. a pre-release like -SNAPSHOT) sorts below the same version without one.
+            .thenComparing(version -> version.qualifier == null);
+
+    private final int major;
+    private final int minor;
+    private final int patch;
+    private final boolean hasMinor;
+    private final boolean hasPatch;
```
mipo256: We can get by without these fields. The semantic version is obligated to have major, minor and patch. There is no other way around. The qualifier is optional. Having these boolean fields is already a sign of a bad design.

common/utils/src/main/java/com/axelixlabs/axelix/common/utils/SemanticVersion.java
```diff
@@ -0,0 +1,228 @@
...
+     */
+    public static Optional<SemanticVersion> tryParse(@Nullable String version) {
+        if (version == null || version.isBlank()) {
+            return Optional.empty();
+        }
+
+        String trimmed = version.trim();
+
+        // Substring the leading major.minor.patch numeric run; everything after it is the qualifier.
+        int[] numbers = {0, 0, 0};
```
mipo256: Do we really like passing mutable array to functions? Why do we do that? This is quite code smell.

Apart from this, returning the array from function and not some nominal type is just a bad idea. It does not signal any intent and it confuses the reader. This is also a code smell

common/utils/src/test/java/com/axelixlabs/axelix/common/utils/SemanticVersionTest.java
```diff
@@ -0,0 +1,254 @@
...
+                    Arguments.of("11", 11),
+                    Arguments.of("2.0", 2),
+                    Arguments.of("1.2.3.4", 1),
+                    Arguments.of("2.5.0-SNAPSHOT", 2),
+                    Arguments.of("3.0.0-RELEASE", 3));
+        }
+    }
+
+    @Nested
+    @TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
mipo256: you fo not need test instance per class here, ditto for all the other places in this test class

mipo256 (changes requested):
We're generally headed to a right direciton. Still, that PR has quite a few design problems

chrshnv:
@coderabbitai review

chrshnv:
@coderabbitai review

chrshnv:
@coderabbitai resume

PR #1294, opened 2026-06-23
Implement Log4j2 support for Spring Boot 3 starter.

sygkanrt30:
Introduce a Log4j2-specific equivalent of the in-memory pagination logging appender by extending AbstractAppender. Register Log4j2InMemoryPaginationAppenderRegistrar conditionally via bean presence validation to achieve mutual exclusivity with the existing Logback configuration. Added integration tests using ApplicationContextRunner to verify runtime precedence of the Logback backend over Log4j2.

Part of #1251

## Summary by CodeRabbit

* **New Features**
  * Added Log4j2 in-memory pagination monitoring with automatic Log4j2 appender registration and detection of the relevant Hibernate in-memory pagination warning.
  * Activation is now correctly gated by the active logging system (Logback vs Log4j2), honors Logback precedence, and can be controlled via `axelix.sbs.transaction.monitoring.in-memory-pagination-detection.enabled` (enabled by default).
* **Tests**
  * Updated/added integration tests to verify Log4j2-vs-Logback precedence and no registration when persistence setup is missing.
* **Chores**
  * Added a compile-time only dependency for Log4j2 starter support.


mipo256 (commented):
There are a couple of nuances when working with auto configurations in Spring Boot. They are not discovered like regular `@Configuration` classes. 

So, the PR would need to adjust slightly to account for them. I will come back to this PR shortly and explain in details.

sygkanrt30:
I've researched this point and I think it's better to extract two top-level configuration classes from the static ones. 

Does this make sense?

sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/TransactionMonitoringAutoConfiguration.java
```diff
@@ -136,9 +138,44 @@ public AxelixMetricsPublisher axelixMetricsPublisher(MeterRegistry meterRegistry
...
 
+        @Bean
+        public LogbackInMemoryPaginationAppenderRegistrar logbackInMemoryPaginationAppenderRegistrar() {
+            return new LogbackInMemoryPaginationAppenderRegistrar();
+        }
+
+        @EventListener(ApplicationReadyEvent.class)
+        public void registerAppender() {
+            logbackInMemoryPaginationAppenderRegistrar().register();
+        }
```
mipo256, line 145: I do not think we need to create a separate bean here. For auto-cofnigruation class, we do not have cglib proxies. 

Right now this is not going to cause problems, but later it suddenly may if some change will sneak into 
`TransactionMonitoringAutoConfiguration`. So as a rule we should generally avoid using self-method invocations. 

```suggestion
        @EventListener(ApplicationReadyEvent.class)
        public void registerAppender() {
            new LogbackInMemoryPaginationAppenderRegistrar().register();
        }
```

sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/TransactionMonitoringAutoConfiguration.java
```diff
@@ -136,9 +138,44 @@ public AxelixMetricsPublisher axelixMetricsPublisher(MeterRegistry meterRegistry
...
+    }
+
+    @Configuration
+    @ConditionalOnHibernateActive
+    @ConditionalOnClass(name = "org.apache.logging.log4j.core.LoggerContext")
+    @AutoConfigureAfter(LogbackInMemoryPaginationAppenderConfiguration.class)
```
mipo256:
```suggestion
```

I think there is no need to do that

mipo256 (changes requested):
So, I finally had some time to outline my thoughts on this @sygkanrt30. I think we're generally on the right track, there is just an issue that needs to be taken care about :)

sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/TransactionMonitoringAutoConfiguration.java
mipo256, line 1: I think we need to revisit the approach here.

We need to have two static configuration classes - one for logback and one for log4j2 (which is partially what we have), and those should be independent (no `@AutoConfigureAfter`) - introducing the dependency between auto configurations is generally should be avoided unless necessary, since it complicates the  overall perseption of the auto-configurations graph.

In the way it is implemented in the PR right now, the auto-configurations is kind of broken. Imagine the case:

People have both logback and log4j2 logger backend in the classpath, but log4j2 is actually being used. Right now, since logback configuration is inspected first, we're just going to misconfigure the listener. This is not a good thing.

What we can do instead is to write a custom conditional annotation (like `@ConditionalOnTheLoggingSystem(system = LOGBACK)`), that inspects the `ILoggingFactory`, and based upon that makes the decision. Only one of those two autoconfigurations will actually boot, and there will be no need to introduce manual ordering between them.

sygkanrt30:
Got it, you're right. Will update.

mipo256 (approved):
This is much better now, there are still a couple of issues to address, but I can polish them myself, or I can leave that up to you. In any case, let me now :)

Thank you for your contribution, @sygkanrt30

sygkanrt30:
What issues have you noticed? Let me know and I'll fix them.  @mipo256

mipo256:
Oh, I have not left them in the review :see_no_evil:. Nevermind then, I'll polish myself.

Thanks for your contribution @sygkanrt30

PR #1295, opened 2026-06-24
GH-573 Refine GC Logging status parsing

### LINKED ISSUE #573: [Feature] Enhance DefaultGcLogService to Properly Detect gc* Logging Configuration and Support Custom Log File Paths
The current DefaultGcLogService implementation has limitations in detecting GC logging status and retrieving log files:

GC Logging Detection Issue: The parseStatus() method only recognizes gc= patterns but fails to detect logging configured with gc*= (wildcard pattern for all GC-related logs). This is a common configuration used for detailed GC monitoring.

Fixed Log File Path: The service assumes GC logs are always written to a default file (gc.log) in the current working directory. It doesn't support custom log file paths specified via Xlog parameters (e.g., file=C:/gclog/gc.log or file=/var/log/myapp/gc.log).

Current Limitations:
Detection: Only recognizes -Xlog:gc=level but not:
1. -Xlog:gc*=level
2. -Xlog:gc+heap=level,gc+pause=level
3. Complex Xlog configurations with output destinations

File Retrieval: Hardcoded to look for gc.log in current directory

Path Parsing: Cannot extract log file path from actual JVM configuration

Expected Behavior:
1. The service should correctly detect GC logging status regardless of the Xlog pattern used (gc=, gc*=, or any GC-related tag combination)
2. The getGcLogFile() method should locate the actual GC log file based on the JVM's current configuration
3. When enabling logging, users should be able to specify custom file paths

Code Analysis:

The parseStatus() method needs enhancement to:
1. Parse various Xlog patterns using regex or more sophisticated string parsing
2. Handle multiple GC-related tags separated by commas
3. Extract log levels from complex configurations

The getGcLogFile() method should:

Query JVM for current logging configuration via jcmd VM.log list

Parse the output to find the actual log file path

Return the correct File object even for custom paths

Test Cases to Consider:
1. -Xlog:gc=info:file=gc.log
2. -Xlog:gc*=debug:file=/var/log/gc.log
3. -Xlog:gc=info:file=C:\\logs\\gc_%t.log
4. -Xlog:gc+heap=info,gc+pause=debug:file=logs/gc.out

sergeycherkasovv:
close #573 

## Summary by CodeRabbit

* **New Features**
  * Added detection of whether JVM GC log output is explicitly configured to write to a file, and surfaced it in GC log insights.

* **Bug Fixes**
  * Improved GC logging enabled detection by better interpreting unified logging selector patterns and deriving the effective level.
  * Fixed GC log status prop wiring in the GC log enable UI so the modal shows the correct current status.

* **Tests**
  * Added deterministic, table-driven tests covering GC log status evaluation and “GC log file specified” detection.

### DIFF
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
@@ -307,8 +307,8 @@ public class DefaultGcLogService implements GcLogService {
      *
      * @see <a href="https://openjdk.org/jeps/158">JEP 158: Unified JVM Logging</a>
      */
-    private boolean isGcSelector(String selector) {
-        String[] parts = selector.split("=", 2);
+    private boolean isGcSelector(String whatSelector) {
+        String[] parts = whatSelector.split("=", 2);
         String tagSet = parts[0];
 
         if ("all".equals(tagSet)) {
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
@@ -64,7 +64,7 @@ public class DefaultGcLogService implements GcLogService {
             ProcessResult result =
                     jcmdExecutor.execute(JCMD_COMMAND, getPid(), JCMD_VM_LOG_COMMAND, JCMD_VM_LOG_LIST_ARGUMENT);
 
-            return parseStatus(result.getOutput(), getAvailableLevels());
+            return parseStatus(result.getOutput());
 
         } catch (Exception e) {
             throw new GcLogException("Failed to get GC log status via jcmd", e);
@@ -265,7 +265,7 @@ public class DefaultGcLogService implements GcLogService {
      *
      * @see <a href="https://openjdk.org/jeps/158">JEP 158: Unified JVM Logging</a>
      */
-    private GcLogStatus parseStatus(String output, List<String> availableLevels) {
+    private GcLogStatus parseStatus(String output) {
         boolean gcSelectorFound = false;
         String highestLevel = null;
 
@@ -295,7 +295,7 @@ public class DefaultGcLogService implements GcLogService {
             }
         }
 
-        return new GcLogStatus(gcSelectorFound, highestLevel, availableLevels);
+        return new GcLogStatus(gcSelectorFound, highestLevel, getAvailableLevels());
     }
 
     /**
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
@@ -20,6 +20,7 @@ package com.axelixlabs.axelix.sbs.spring.core.gclog;
 import java.io.File;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Locale;
 import java.util.stream.Collectors;
 
 import org.jspecify.annotations.Nullable;
@@ -32,10 +33,16 @@ import com.axelixlabs.axelix.sbs.spring.core.log.Logger;
  *
  * @since 30.12.2025
  * @author Nikita Kirillov
+ * @author Sergey Cherkasov
  */
 public class DefaultGcLogService implements GcLogService {
 
+    private static final String JCMD_COMMAND = "jcmd";
+    private static final String JCMD_VM_LOG_COMMAND = "VM.log";
+    private static final String JCMD_VM_LOG_LIST_ARGUMENT = "list";
     private static final String DEFAULT_FILE_NAME = "gc.log";
+    private static final String DEFAULT_LOG_LEVEL = "info";
+    private static final String OFF_LOG_LEVEL = "off";
 
     private final JcmdExecutor jcmdExecutor;
     private final Logger logger;
@@ -51,19 +58,21 @@ public class DefaultGcLogService implements GcLogService {
         this.logger = logger;
     }
 
-    // TODO We need to enhance the GC Logging status monitoring https://github.com/axelixlabs/axelix/issues/573
     @Override
     public GcLogStatus getStatus() {
         try {
-            ProcessResult result = jcmdExecutor.execute("jcmd", getPid(), "VM.log", "list");
+            ProcessResult result =
+                    jcmdExecutor.execute(JCMD_COMMAND, getPid(), JCMD_VM_LOG_COMMAND, JCMD_VM_LOG_LIST_ARGUMENT);
 
-            return parseStatus(result.getOutput());
+            return parseStatus(result.getOutput(), getAvailableLevels());
 
         } catch (Exception e) {
             throw new GcLogException("Failed to get GC log status via jcmd", e);
         }
     }
 
+    // TODO: This assumes that the GC log file is always gc.log.
+    //       In reality, the file name and path can be customized, e.g. file=/tmp/gc.log.
     @Override
     public File getGcLogFile() throws GcLogException {
         File file = new File(DEFAULT_FILE_NAME);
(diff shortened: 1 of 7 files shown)


sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -193,33 +214,86 @@ private List<String> loadAvailableLevels() {
...
+                String[] parts = selector.split("=", 2);
+                String level = parts.length == 2 ? parts[1] : DEFAULT_LOG_LEVEL;
+                if (!OFF_LOG_LEVEL.equals(level) && verbosity(level) > verbosity(highestLevel)) {
+                    highestLevel = level;
+                }
             }
         }
 
-        return new GcLogStatus(false, null, getAvailableLevels());
+        return new GcLogStatus(gcSelectorFound, highestLevel, availableLevels);
```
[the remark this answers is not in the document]
sergeycherkasovv: This is not entirely correct and is taken out of context; before line 252 a check is performed, and if it passes, then `gcSelectorFound = true;` is applied.
```java
if (!isGcSelector(selector)) {
    continue;
}
```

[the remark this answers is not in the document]
sergeycherkasovv: We aim to cover a large number of scenarios, and the case where someone simply sets `gc=off` without enabling other tags is unlikely, and 90% of users won’t do that

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -75,6 +78,38 @@ public File getGcLogFile() throws GcLogException {
...
+                // Expected format: "#0: <output> <what> <decorators> ...", where <what> contains log selectors.
+                String[] configurationParts = trim.split("\\s+");
+                if (configurationParts.length < 3 || !configurationParts[1].startsWith("file=")) {
+                    continue;
+                }
+
+                String what = configurationParts[2];
+                for (String selectorText : what.split(",")) {
+                    if (isGcSelector(selectorText.trim())) {
+                        return true;
```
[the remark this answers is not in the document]
sergeycherkasovv: We aim to cover a large number of scenarios, and the case where someone simply sets gc=off without enabling other tags is unlikely, and 90% of users won’t do that.

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -193,33 +228,86 @@ private List<String> loadAvailableLevels() {
...
+            case "info":
+                return 2;
+            case "debug":
+                return 3;
+            case "trace":
+                return 4;
+            default:
+                return -1;
         }
     }
```
sergeycherkasovv: In the [JEP 158 documentation](https://openjdk.org/jeps/158?utm_source=chatgpt.com), there is the phrase **“The available levels are error, warning, info, debug, trace and develop in increasing order of verbosity.”**, which, as I understand it, implies that each subsequent level includes the previous ones, that is, `trace` also shows `debug`, `info`, `warning`, and `error`.

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -193,33 +228,86 @@ private List<String> loadAvailableLevels() {
...
+    private boolean isGcSelector(String selector) {
+        String[] parts = selector.split("=", 2);
+        String tagSet = parts[0];
+
+        if ("all".equals(tagSet)) {
+            String level = parts.length == 2 ? parts[1] : DEFAULT_LOG_LEVEL;
+            // JVM default logging uses all=warning, which technically includes GC warning/error messages.
+            // For the product status, treat broad all selectors as GC logging only at info/debug/trace.
+            return verbosity(level) >= verbosity(DEFAULT_LOG_LEVEL);
         }
```
sergeycherkasovv, line 317: It’s important to note that we exclude levels such as `error` and `warning` from the all `tag`, because they are enabled by default when the application starts. Therefore, `all=warning` does not necessarily mean that **GC Logging** was explicitly enabled by the user.

[JEP 158](https://openjdk.org/jeps/158?utm_source=chatgpt.com)

```
Default configuration:
-Xlog:all=warning:stderr:uptime,level,tags
    - default configuration if nothing is configured on command line
    - 'all' is a special tag name aliasing all existing tags
    - this configuration will log all messages with a level that
    matches ´warning´ or ´error´ regardless of what tags the
    message is associated with
```

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -75,6 +78,38 @@ public File getGcLogFile() throws GcLogException {
         return file;
     }
 
+    @Override
+    public boolean isGcLogFileSpecified() throws GcLogException {
+        try {
+            ProcessResult result = jcmdExecutor.execute("jcmd", getPid(), "VM.log", "list");
```
sergeycherkasovv: I’m not saying this is something we should necessarily do, but it seems that `jcmd` could potentially be replaced with [DiagnosticCommandMBean](https://docs.oracle.com/en/java/javase/17/docs/api/jdk.management/com/sun/management/DiagnosticCommandMBean.html). At least for `VM.log list`, it may provide the same diagnostic command output without spawning an external `jcmd` process.

It also looks like this approach could allow us to enable or disable GC Logging and retrieve the configured log file, since these operations are exposed through the same `VM.log` diagnostic command.

However, this would not remove the need for parsing. The output would still be the textual `VM.log list` result, so we would still need to analyze it in the same way.

NikitaKirilloff (changes requested):
Following up on our review and online discussion, here are the main conclusions we reached:

1. At the developer's discretion. Add detailed comments in the code logic where it is truly needed.

2. We need to rebase onto master and test the GC logging detection logic across various scenarios using our playground projects.

3. We must verify if we accurately determine the filename. We need to be absolutely certain that this file contains GC logs before sending it to our UI. This might require additional adjustments and should likely be handled in a separate PR.

4. We should explore moving from jcmd to DiagnosticCommandMBean in a future PR. Since DiagnosticCommandMBean is HotSpot-specific, we might need a custom condition like OnHotSpotCondition (or something similar), though this is just an initial assumption for now.

5. Let's consider dropping the GcLogStatus#level field, as it is currently only used in tests. We need to evaluate whether we actually need it at all, which requires further investigation and discussion with the team.

@mipo256 @sergeycherkasovv

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
sergeycherkasovv: @coderabbitai create an issue to We need to stop using `jcmd` and instead use the **HotSpot** `DiagnosticCommandMBean` tool

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -193,33 +228,86 @@ private List<String> loadAvailableLevels() {
...
                 }
 
-                String level = trim.substring(idx + 3, end);
-                return new GcLogStatus(true, level, getAvailableLevels());
+                gcSelectorFound = true;
+                String[] parts = selector.split("=", 2);
+                String level = parts.length == 2 ? parts[1] : DEFAULT_LOG_LEVEL;
+                if (!OFF_LOG_LEVEL.equals(level) && verbosity(level) > verbosity(highestLevel)) {
+                    highestLevel = level;
+                }
```
sergeycherkasovv, line 294: After a recent call with @NikitaKirilloff , we concluded confirmed by his [comment](https://github.com/axelixlabs/axelix/pull/1295#pullrequestreview-4624295049) that the `GcLogStatus#level` field is not used anywhere in our application logic. However, we also reached another conclusion: in my implementation, a combined **GC Logging** configuration like `gc=info,gc+pause=debug` would result in `debug`, since it includes more detailed levels than `info`. This is not correct, because in this case there is no single definitive logging level for GC. I believe this is a rare but valid scenario

Perhaps we should remove `GcLogStatus#level`

sbs/starter-domain/src/test/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogServiceTest.java
```diff
@@ -81,6 +83,144 @@ private static Stream<String> availableLevelsProvider() {
         return subject.getStatus().getAvailableLevels().stream();
     }
 
+    @ParameterizedTest
+    @MethodSource("enabledGcLogStatusProvider")
+    void getStatus_shouldReturnEnabled_whenUnifiedLoggingSelectorsContainGc(String vmLogOutput, String level) {
+        // given.
+        var subject = new DefaultGcLogService(new StaticJcmdExecutor(vmLogOutput), new NoOpLogger());
```
sergeycherkasovv, line 90: This was done intentionally to avoid invoking the `jcmd` process every time. After we stop using `jcmd` and switch to `DiagnosticCommandMBean` in this [task](https://github.com/axelixlabs/axelix/issues/1367), this test class will require minor changes to be rewritten to use `DiagnosticCommandMBean`)

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
```diff
@@ -51,19 +58,21 @@ public DefaultGcLogService(JcmdExecutor jcmdExecutor, Logger logger) {
...
+            ProcessResult result =
+                    jcmdExecutor.execute(JCMD_COMMAND, getPid(), JCMD_VM_LOG_COMMAND, JCMD_VM_LOG_LIST_ARGUMENT);
 
-            return parseStatus(result.getOutput());
+            return parseStatus(result.getOutput(), getAvailableLevels());
 
         } catch (Exception e) {
             throw new GcLogException("Failed to get GC log status via jcmd", e);
         }
     }
```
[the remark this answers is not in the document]
sergeycherkasovv: On this stage, there is no need to create a cache for two reasons:
1. There is an ability to modify the `VM.log list` in Runtime
2. We are planning to stop using `jcmd` in favor of `DiagnosticCommandMBean`

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/gclog/DefaultGcLogService.java
mipo256: I think we need to give it more thoughts. Generally, the root of the problem is just we assume that there is a single logging level for the entire GC process, which is incorrect.

And thus the UI that currently provides a way to change the logging level of gc is also at least a bit delusive.

mipo256 (approved):

PR #1297, opened 2026-06-24
GH-1004: Super admin bcrypt support

### LINKED ISSUE #1004: [Feature] Accept hashed passwords for SUPER_ADMIN
Add the possibility to accept the hashed password for the `SUPER_ADMIN` from the configuration sources (like bcrypt hash or something). We must be able to accept both plain and hashed passwords.

reptiloyds:
Closes #1004.

## What changed
- Added `SuperAdminPasswordEncoder` to support `plaintext`, `{noop}` and `{bcrypt}` password formats for the super-admin credentials.
- Modified `SuperAdminUserAuthenticator` to validate password format at startup (`@PostConstruct`) and use `SuperAdminPasswordEncoder` for matching.
- Added unit tests for `SuperAdminPasswordEncoder` (validation, matching, extraction) and `SuperAdminUserAuthenticator`.
- Updated documentation to describe supported password formats for the super-admin.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved super-admin login by validating the configured password format at startup and verifying credentials using secure password matching (no direct text comparison).
  * Added support for `{id}encodedValue` password prefixes (e.g., `{bcrypt}`, `{noop}`); if no prefix is provided, the value is treated as plaintext.
  * Unsupported or malformed password formats are now rejected early.
* **Documentation**
  * Updated authentication setup guides (including Russian) to describe supported password formats and recommend hashed values like `{bcrypt}`.
* **Tests**
  * Added coverage for password format validation, matching, and prefix extraction behavior.


docs/docs/setting-up-master-ui/authentication/authentication.mdx
```diff
@@ -30,13 +30,22 @@ local users, followed by a single **Sign in with OAuth2/OIDC Provider** button w
...
 sign-in grants the `SUPER_ADMIN` role, which carries every authority Master defines (see [Roles and
 authorities](#roles-and-authorities)).
 
 This account is meant to bootstrap Master and to recover access when other providers are misconfigured. Override the
 defaults before exposing Master to anyone — `admin / admin` grants full control over every monitored service.
 
+The password can be provided in one of the following formats:
+
+* **Plaintext**
+* **[DelegatingPasswordEncoder](https://docs.spring.io/spring-security/reference/api/java/org/springframework/security/crypto/password/DelegatingPasswordEncoder.html)** — `{noop}` or `{bcrypt}` prefixes.
```
mipo256: I would rephrase it, since the system administrators that are going to deploy Axelix are probably not aware of `DelegatingPasswordEncoder` abstraction in the Spring Framework. I would at least elaborate a bit on what that actually means.

docs/i18n/ru/docusaurus-plugin-content-docs/current/setting-up-master-ui/authentication/authentication.mdx
```diff
@@ -34,13 +34,22 @@ Provider»** при включённом OIDC.
...
 которая даёт все полномочия, определённые в Master (см. [Роли и полномочия](#roles-and-authorities)).
 
 Эта учётная запись предназначена для начальной настройки Master и восстановления доступа в случае неправильной настройки
 других провайдеров. Прежде чем предоставлять доступ к Master кому бы то ни было, измените настройки по умолчанию —
 `admin / admin` даёт полный контроль над каждым мониторируемым сервисом.
 
+Пароль может быть указан в одном из следующих форматов:
+
+* **Plaintext**
+* **[DelegatingPasswordEncoder](https://docs.spring.io/spring-security/reference/api/java/org/springframework/security/crypto/password/DelegatingPasswordEncoder.html)** — с префиксами {noop} или {bcrypt}.
```
mipo256: Ditto

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/auth/properties/SuperAdminConfigurationProperties.java
```diff
@@ -55,5 +55,10 @@ public String getPassword() {
         return credentials.password();
     }
 
+    /**
+     * @param username super-admin username
+     * @param password plain text or encoded password using DelegatingPasswordEncoder format
+     * {@code {noop}}, {@code {bcrypt}}
+     */
```
mipo256, line 63: Add yourself as the author, please.

master/src/main/java/com/axelixlabs/axelix/master/service/auth/encoder/SuperAdminPasswordEncoder.java
```diff
@@ -0,0 +1,145 @@
...
+        if (isUnsupportedEncoderId(encoderId)) {
+            throw new IllegalArgumentException("The " + SUPER_ADMIN_LOGIN_PROPERTIES_PREFIX
+                    + ".credentials.password uses the encoder id " + encoderId
+                    + " which is unsupported.");
+        }
+
+        if (isNotHashedFormat(encoderId)) {
+            log.warn(
+                    "The {}.credentials.password is not hashed. Consider storing the password using a supported DelegatingPasswordEncoder format (e.g. {bcrypt}).",
+                    SUPER_ADMIN_LOGIN_PROPERTIES_PREFIX);
```
mipo256:
```suggestion
            log.warn(
                    "The " + SUPER_ADMIN_LOGIN_PROPERTIES_PREFIX + ".credentials.password is not hashed. Consider storing the password using a supported DelegatingPasswordEncoder format (e.g. {bcrypt}).");
```

I think there is no need to defer the string concatenation here. Let's stick to the common pattern above.

master/src/test/java/com/axelixlabs/axelix/master/service/auth/encoder/SuperAdminPasswordEncoderTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final String PLAIN_PASSWORD = "password";
+
+    @Spy
+    private BCryptPasswordEncoder bcryptPasswordEncoder = new BCryptPasswordEncoder();
+
+    @InjectMocks
+    private SuperAdminPasswordEncoder encoder;
+
+    @Nested
+    class ValidationTests {
```
mipo256, line 53: We can reduce the number of test methods here from 6 to 3 by using parametrized junit tests. I think it is worth to refactor it.

master/src/test/java/com/axelixlabs/axelix/master/service/auth/encoder/SuperAdminPasswordEncoderTest.java
```diff
@@ -0,0 +1,167 @@
...
+ *
+ * @author Ilya Naumov
+ */
+@ExtendWith(MockitoExtension.class)
+@ExtendWith(OutputCaptureExtension.class)
+public class SuperAdminPasswordEncoderTest {
+    private static final String PLAIN_PASSWORD = "password";
+
+    @Spy
+    private BCryptPasswordEncoder bcryptPasswordEncoder = new BCryptPasswordEncoder();
```
mipo256: Do we need to make it a spy and do we need the `MockitoExtension`  at all? 

I mean, AFAICT we can make it way simpler by just:

```
@ExtendWith(OutputCaptureExtension.class)
public class SuperAdminPasswordEncoderTest {

    private static final String PLAIN_PASSWORD = "password";

    private SuperAdminPasswordEncoder encoder;

    @BeforeEach
    void setup() {
         this.encoder = new SuperAdminPasswordEncoder(new BCryptPasswordEncoder());
    }
```

And that seems to be just it

master/src/test/java/com/axelixlabs/axelix/master/service/auth/encoder/SuperAdminPasswordEncoderTest.java
```diff
@@ -0,0 +1,167 @@
...
+            // when.
+            assertThatThrownBy(callable)
+                    // then.
+                    .isInstanceOf(IllegalArgumentException.class);
+        }
+    }
+
+    @Nested
+    class MatchTests {
+        @Test // GH-1004
```
mipo256: I would also instead introduce two parametrized tests, one for matches and the other for non-matches cases. I think this way we're going to cover more cases (empty string, noop prefix, no prefix at all etc) than just doing this.

master/src/test/java/com/axelixlabs/axelix/master/service/auth/provider/SuperAdminUserAuthenticatorTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Mock
+    private SuperAdminConfigurationProperties superAdminConfiguration;
+
+    @InjectMocks
+    private SuperAdminUserAuthenticator authenticator;
+
+    @Nested
+    class ValidationTests {
+        @Test // GH-1004
+        void shouldInvokePasswordEncoderValidation() {
```
mipo256: I think this is the wrong approach overall.

We need to test not just the fact that `validate` works correctly, but the fact that it fails the context launch if the provided password is incorrect. We should not use mockito here, we should use `ApplicationContextRunner`.

master/src/test/java/com/axelixlabs/axelix/master/service/auth/provider/SuperAdminUserAuthenticatorTest.java
```diff
@@ -0,0 +1,125 @@
...
+            // when.
+            authenticator.validate();
+
+            // then.
+            verify(encoder, times(1)).validatePasswordFormat(PLAIN_PASSWORD);
+        }
+    }
+
+    @Nested
+    class AuthenticateTests {
```
mipo256: I think we should not use Mockito here, we should instead use `@SpringBootTest` tests to really make sure the binding actually happens the way we expect. Right now the quality of the test is not that good, and it also adds up quite a lot of `when` mockito noise.

mipo256 (changes requested):
I think we're on the right track, this change looks reasonably good. Still, a couple of issues needs to be resolved

reptiloyds:
I think I've resolved the reported issues. However, I've now hit the pull request size limit and I'm not entirely sure how to proceed in this case. Could you advise on the preferred approach? @mipo256

mipo256 (approved):
I think overall we're good. Thank you for your contribution, @reptiloyds !

PR #1306, opened 2026-06-26
GH-1292: Add instance name postfix generation

### LINKED ISSUE #1292: [BUG] Ensure uniqueness of Instance names during self-registration
Right now, if people deploy Axelix with self-registration as the primary discovery mechanism, and if there are multplie instances of the managed services deployed, all of these same instances will share the same `Instance#name`, which is quite bad.

**TBD:**

We need to find the way to mangle the name of the Instance a bit, to make it unique, during self-registration. I am thinking about adding a small, maybe around 8 letters long ASCII randomly-generated postfix.

reptiloyds:
Closes #1292.

## What changed
- Added random 8-char postfix (lowercase letters + digits) to instance name in `DefaultSelfRegistrationMetadataAssembler` to ensure unique instance names.
- Updated `DefaultSelfRegistrationMetadataAssemblerTest` - changed assertion to verify instance name starts with configured name + 8-char postfix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Registration metadata instance names now include a generated unique suffix in addition to the configured base name, reducing collisions when multiple instances register.
  * Metadata creation now consistently uses this full generated instance name.

* **Tests**
  * Updated assertions to match the new instance-name format (prefix plus expected length).
  * Added coverage to confirm name stability within repeated assemblies and uniqueness across different assembler instances.

### DIFF
--- a/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultSelfRegistrationMetadataAssembler.java
+++ b/sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultSelfRegistrationMetadataAssembler.java
@@ -18,6 +18,7 @@
 package com.axelixlabs.axelix.sbs.spring.core.master;
 
 import java.time.Instant;
+import java.util.Random;
 import java.util.UUID;
 
 import com.axelixlabs.axelix.common.api.registration.SelfRegistrationMetadata;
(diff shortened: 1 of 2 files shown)


mipo256 (changes requested):
The direction overall is good, still, there is a problem.

The name must be stable. The metadata assembler is called continuously during the self-registration heartbeating. This implementation is just going to regenerate the name which is wrong - the name must be stable. I think we can just calculate it first and then cache it.

We should add the test case for that in the `DefaultSelfRegistrationMetadataAssembler`, so that we do not regenerate the name on the repetitive calls.

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/master/DefaultSelfRegistrationMetadataAssembler.java
```diff
@@ -45,16 +54,26 @@ public DefaultSelfRegistrationMetadataAssembler(
         this.selfRegistrationConfigurationProperties = selfRegistrationConfigurationProperties;
         this.serviceMetadataAssembler = serviceMetadataAssembler;
         this.instanceId = UUID.randomUUID().toString();
+        this.instanceName = selfRegistrationConfigurationProperties.getInstanceName() + "-" + generateNamePostfix();
```
mipo256, line 57: I'd use `Lazy` here, but that is already nitpicking.

mipo256 (approved):
We're generally good. Thanks for the patch @reptiloyds . I also think it is worth outlining this behavior in docs, since it may be unclear on why `instanceName` has the postfix. But I'm going to do that myself.

PR #1319, opened 2026-07-01
GH-1318 Disallow overriding beans in `AxelixBeansEndpointAutoConfiguration` and `AxelixConditionsEndpointAutoConfiguration`

### LINKED ISSUE #1318: [Feature] Disallow overriding beans in `AxelixBeansEndpointAutoConfiguration` and `AxelixConditionsEndpointAutoConfiguration`

sergeycherkasovv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-configured beans for the enabled endpoints now register consistently across Spring Boot 2 and Spring Boot 3.
  * If an application supplies conflicting beans for these endpoint components, startup now fails fast with a bean-definition override error instead of silently changing wiring.
* **Tests**
  * Updated autoconfiguration tests to expect startup failure (including bean-definition override assertions) when custom overriding beans are provided.


mipo256 (changes requested):
Overall okay, I just think that there is no need to change tests that used to verify, that the custom beans were picked up - we can simply delete these tests.

The idea is simple - we will reduce the amount of code in the repo, which is good. And moreover - when people provide their own beans they are entering an undocumented field. We are not promising here that we will fail with exception. There is no need to check anything here.

mipo256 (approved):

mipo256 (approved):

PR #1331, opened 2026-07-02
(sbs3) Support ignoring architecture tests with @IgnoreArchitectureTest annotation

chrshnv:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated several test suites so they’re excluded from architecture checks, helping prevent unrelated test failures and improving overall build reliability.
* **Chores**
  * Added a new test-only annotation to make these exclusions easier to apply consistently.


mipo256 (approved):
I generally think we're good, although, I think, it will be a good idea to make a mandatory `reason` attribute for the annotation, so that people think twice before applying it.

PR #1332, opened 2026-07-02
(sbs3) test: add test for children test classes that shouldn't be annotated with restricted annotations

chrshnv:
## Summary by CodeRabbit

* **Tests**
  * Strengthened automated architecture checks to prevent certain Spring test-context annotations from being applied to the wrong kinds of test classes within the core package.
  * Added additional ArchUnit coverage to validate nested classes and nested methods are annotated according to the allowed set.
  * Improved rule clarity by standardizing the restricted-annotations constant name, keeping behavior consistent.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/TestContextArchitectureTest.java
+++ b/sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/TestContextArchitectureTest.java
@@ -20,6 +20,16 @@ package com.axelixlabs.axelix.sbs.spring.core;
 import java.lang.annotation.Annotation;
 import java.util.List;
 
+import com.tngtech.archunit.base.DescribedPredicate;
+import com.tngtech.archunit.core.domain.JavaAnnotation;
+import com.tngtech.archunit.core.domain.JavaClass;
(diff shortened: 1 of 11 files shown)


sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/TestContextArchitectureTest.java
```diff
@@ -131,4 +147,53 @@ public class TestContextArchitectureTest {
             AutoConfigureRestDocs.class,
             WebServiceClientTest.class,
             ImportAutoConfiguration.class);
+
+    public static final JavaClasses IMPORTED_CLASSES =
+            new ClassFileImporter().importPackages("com.axelixlabs.axelix.sbs.spring.core");
+
+    @Test
+    void should_have_restricted_annotations_only_on_parent() {
```
mipo256: Do you really think we need this rule? I don't, TBH. This rule is going to be fragile, and it is unclear why we even need it. If we're unsure, then lets remove it.

Instead, what we **do** need is the rule that enforces that the subclasses of the abstract parent context classes are not introducing any new context themselves.

mipo256 (changes requested):
Please, rebase this pr onto master

mipo256 (approved):

Issue #1314, opened 2026-06-29
[Feature] Make sure the CodeRabbit can access skills during the review


mipo256:
As of now, our skills are stored in the `.agents_skills`, and the particular directory layouts are not committed to VCS and are gitignored with the goal of not repeating the configurations. 

Because of this, CodeRabbit is unable to see the skills during the review rounds, and this is very bad since we have a lot of contextual knowledge here, in particular  `config-breaking-changes` skill.

**DOD:**

CodeRabbit is able to access skills during the review.

Issue #1155, opened 2026-06-04
[Feature] Improve the backward compatibility on starters side


mipo256:
Currently, in starters, we have a lot of stuff that is really internal, marked as `public`. That is going to cost us a maintenance burden in the future. So, we should consider changing the visibility of the API.

I guess we can also write a skill for reviewing the code that modifies the starters. The main objectives are:

1. Resist making the API public unless it is really intended (which it is almost never is). If we can make it internal/hide it -  we should do so.

2. Apart from that, CodeRabbit / AI Agent must add the label `breaking-changes` if we're changing the public api

Issue #812, opened 2026-03-08
[Refactoring] Revisit package naming for sbs modules


mipo256:
Currently, it is impossible to differentiate classes in `starter-domain`, `spring-boot-2-starter` and `spring-boot-3-starter` the code base.

This circumstance interferes with the developer's short-term context window, as he/she cannot deduce from the class definition alone where exactly this class is located. 

**IMPORTANT** 

The vast majority of PRs affecting the starter must be merged before we start muggling with package names.

PR #1362, opened 2026-07-05
Add automatic bean renaming processor to avoid name collisions.

### LINKED ISSUE #957: [Feature]  Standardize bean naming strategy to prevent collisions
We need to revise our bean naming convention within the axelix sbs. Currently, names like authorizer, securityManager, jwtDecoderService, and jwtEncoderService are too generic. They risk colliding with standard Spring Security beans or custom beans defined in client applications.

Proposed Solution: Consider implementing a consistent prefixing rule. We could introduce a custom annotation, such as @AxelixBean, to ensure all starter-provided beans are automatically prefixed with axelix (e.g., axelixAuthorizer, axelixSecurityManager).

sygkanrt30:
Implement AxelixBeanRenamingProcessor to automatically apply the "axelix" prefix to all internal beans created via @Bean methods. The implementation safely copies the bean names array to prevent ConcurrentModificationException during registry modification.

Closes #957

## Summary by CodeRabbit

* **New Features**
  * Added Spring Boot auto-configuration for both starter variants to enable automatic bean name renaming to the `axelix` convention.
* **Bug Fixes**
  * Prevented beans that already follow the convention from being prefixed twice.
  * Ensured renamed beans remain resolvable via aliases so lookups keep working reliably.
* **Tests**
  * Added integration tests validating both the renaming behavior and the already-compliant no-change scenario.

### DIFF
--- a/sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/AxelixCoreNamingAutoConfiguration.java
+++ b/sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/autoconfiguration/AxelixCoreNamingAutoConfiguration.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.sbs.spring.autoconfiguration;
+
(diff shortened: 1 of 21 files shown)


mipo256 (changes requested):
The overall approach to the problem is good @sygkanrt30. Still, there is one critical issue.

Currently we have not yet configured CodeRabbit to correctly parse custom skills in the `.agent_skill`, we have the ticket for that (#1314), so CodeRabbit did not notice the important architectural problem: **we cannot add any spring-related capabilities to the domain module.** 

That module is expected to be independent from the Spring Boot version, and thus it should be shared across all starters. It therefore cannot contain spring-related facilities. 

The solution would be to have different factory processors for different starters. They are probably going to be very similar in boot 2 and 3.

mipo256:
I am going to review this PR shortly. Can you please revisit the tests? It seems that some of our tests were in fact dependent on the actual bean name (which is wrong). 

I can do that myself later in post-merge polishing, but I would really appreciate seeing that in the PR.

In any case, please, let me know :wink: :)

sygkanrt30:
Yep, I will try to fix it. @mipo256

sygkanrt30:
Hey, what if we scope the package down to autoconfigs in processor? WDYT? @mipo256 This aligns with the original ticket requirement and fixes the issue with the broken tests.

mipo256:
> Hey, what if we scope the package down to autoconfigs in processor? WDYT? @mipo256 This aligns with the original ticket requirement and fixes the issue with the broken tests.

I think that this is going to be a bit fragile to depend upon the packages' structure. We're going to change that in starters soon, multiple times BTW (see #1155 and #812). The dependency upon the packages structure is unclear and it is too implicit. We have already shot ourselves multiplie times with it.

Can we avoid that @sygkanrt30 ?

sygkanrt30:
Sure, we can!

mipo256 (approved):
I've made some minor polishing, other than that - the PR is pretty good, thanks mate @sygkanrt30 !

PR #1421, opened 2026-07-13
GH-1216: Copy spring test profiler reports to application classpath

### LINKED ISSUE #1216: [Feature] Creating Gradle plugin for Axelix

chrshnv:
Closes #1216

## Summary by CodeRabbit

* **New Features**
  * Added configurable control over profiler report copying (enabled by default).
  * Automatically generates and packages build metadata (`META-INF/axelix-info.properties`), including Git details when available.
  * Improves Java/Gradle compatibility so profiler support and report behavior adapt to the project’s target Java version.
* **Bug Fixes**
  * Avoids adding unnecessary profiler/Thymeleaf entries and preserves existing `spring.factories` registrations.
  * Ensures the profiler report is copied before `bootJar` when applicable.
* **Tests**
  * Added functional coverage for profiler wiring, report-copy behavior, and project-info generation.


chrshnv:
@coderabbitai review

Issue #1219, opened 2026-06-13
[Feature] Add support for detecting the layered docker images being used as a feature


mipo256:
Introduce support for detecting [the layered docker images](https://docs.spring.io/spring-boot/reference/packaging/container-images/efficient-images.html#packaging.container-images.efficient-images.layering) being used as a feature.

mipo256:
## Suggested implementation approach

This can be implemented as a new Spring Framework insight feature (alongside the existing ones like `OSIV`), surfaced through `BasicDiscoveryMetadata.Insights` → `springFramework` and populated from the starter side in `DefaultInsightsInfoProvider` (or a small dedicated detector in `starter-domain`).

The goal is to detect whether the application is running in the **extracted / layered Docker image layout** described in the [efficient container images](https://docs.spring.io/spring-boot/reference/packaging/container-images/efficient-images.html) and [Dockerfiles](https://docs.spring.io/spring-boot/reference/packaging/container-images/dockerfiles.html) docs — i.e. the runtime is *not* the original uber jar, but a thin application jar that references dependencies extracted to the filesystem.

### Detection signals

**1. Class loader type (strong negative signal for layered layout)**

When Spring Boot runs a classic fat jar (`java -jar my-app.jar` on the uber jar), dependencies under `BOOT-INF/lib/` are loaded via Spring Boot's custom `URLClassLoader` hierarchy — historically `org.springframework.boot.loader.LaunchedURLClassLoader`, in newer Boot versions `org.springframework.boot.loader.launch.LaunchedURLClassLoader`.

If the thread context / application class loader chain contains one of these Boot loader classes, we can treat that as a **definitive indicator that the app is running from the uber jar** (nested-jar layout). In that case the layered-docker-image feature should be reported as **disabled**.

This is a reliable shortcut: custom Boot `URLClassLoader` ⇒ 100% uber-jar mode.

**2. `Class-Path` manifest attribute (positive signal for layered layout)**

In the efficient Docker image workflow (`jarmode extract --layers`), the runtime `application.jar` is intentionally *not* the builder uber jar. Spring Boot documents that it *"only contains application code and references to the extracted jar files"*. Those references are carried in the jar manifest's `Class-Path` attribute, pointing at the dependency jars that were copied into separate image layers.

At runtime we can read the manifest of the application's code source (e.g. via `getProtectionDomain().getCodeSource().getLocation()` on an application class) and check whether `Class-Path` is present and non-empty.

[...]

PR #1427, opened 2026-07-13
GH-1219: Add JAR manifest inspection and JarLauncher detection.

reptiloyds:
Part of #1219.

## Summary
Foundational changes for **GH-1219**. Introduces interfaces and implementations for detecting JAR manifest attributes and Spring Boot `JarLauncher` usage. These components will be wired into `LayeredDockerImageStateProvider` in subsequent pull requests.

##  What changed
- Added `warn()` to `Logger` interface + implementations in sbs2/sbs3.
- New `JarLauncherDetector` and `JarManifestInspector` interfaces.
- `DefaultJarLauncherDetector` and `DefaultJarManifestInspector` with tests.

## Summary by CodeRabbit

* **New Features**
  * Added automatic detection of whether the application is running in Docker or Kubernetes environments.
  * Supports configurable Docker marker locations for flexible runtime detection.

* **Tests**
  * Added coverage for Docker, Kubernetes, and combined container-detection scenarios, including positive and negative cases.


mipo256 (changes requested):
So, a couple of things. I have spent some time thinking of **what exactly is the thing we're trying to detect**, what is the value for the end-user.

The idea is that we want to target the fact that end-user uses the extracted jar layout with the regular classloader, not the boot's `LaucnherURLClassLoader`.

So the algorithm must be quite simple. We assume that the end user is using the effective docker images if and only if:

1. The classloader that used to load our library class (like any class from our `starter-domain`) is NOT `LaucnherURLClassLoader` or its new analog. **Not the context ClassLoader**. Context ClassLoader is not in our applicable. Let's not mix things around

2. And what is even more important, if our starter as a jar does NOT reside inside the fat jar file system.

**Of course, all of the above only makes sense if we detect the Docker environment.** Axelix supports bare metal deployment of JAR files or deployment of JARs on VMs with no docker. So if we do not detect the Docker environment, then effective docker images are not applicable, obviously. 

if **ANY** of the above are false, and we're running inside the docker layout, then we can suggest the user to use effective docker iamges.

reptiloyds:
Okay, I understand the idea behind the new approach.

However, I have a question about the necessity of checking for `LaunchedURLClassLoader / LaunchedClassLoader`. It doesn't seem like a reliable indicator to me.

I did some research and built several test images. **If I'm not mistaken**, checking for `LaunchedClassLoader` will return true for both a fat JAR and an extracted/layered layout if the user builds the image using:
`java -Djarmode=tools -jar app.jar extract --layers --launcher`
and starts the container via `org.springframework.boot.loader.launch.JarLauncher`.

Moreover, for Spring Boot 2, launching a layered container through `JarLauncher` is actually the default scenario.

This is why the algorithm I was working on also checked the URL protocol of app location whenever a `LaunchedClassLoader` was detected.

Given that, would it make more sense for the algorithm to simply check whether we're running in a Docker environment and then verify the URL protocol instead?

mipo256:
> I did some research and built several test images. If I'm not mistaken, checking for LaunchedClassLoader will return true for both a fat JAR and an extracted/layered layout if the user builds the image using:
java -Djarmode=tools -jar app.jar extract --layers --launcher
and starts the container via org.springframework.boot.loader.launch.JarLauncher.

Alright, that generally makes sense to my. The problem we want to solve in the first place is to ensure that people use layared images. If our library is living outside the jar then I guess we can make the conclusion already. You're probably right. 

That would also allow for having htis entire logic be posiitoned in `starter-domain`. Let's give it a try.

reptiloyds:
At this point, I've added `ContainerEnvironmentDetector`.

While implementing it, I started wondering whether we should be trying to detect only Docker. AFAIU, layered image builds are generally relevant across the OCI, not just Docker.

For that reason, I’m planning to add another signal as a fallback when the primary checks do not match. In addition to the obvious scenarios already covered (K8S and Docker), we could inspect PID 1 and check whether it is not one of the traditional init processes (`init` or `systemd`). In theory, this could help cover other OCI implementations, such as Podman.

What do you think about this approach?

PR #1434, opened 2026-07-15
Create plugin, which generate auto config imports file, add it to sbs 2 and 3

### LINKED ISSUE #722: [Feature] Generate Spring Boot AutoConfiguraton.imports metadata file programmatically
When reviewing #670, it became clear that as the number of the auto-configuration classes grows, we cannot rely on the end developer to remember to add the AutoConfiguration entry in the required spring boot `META-INF` config file.

I think we can write a plugin to auto-generate this file based on the structure of the autoconfiguration package.

sygkanrt30:
Write AxelixAutoConfigPlugin, which scan code source and generate "META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports" consist of class full names, if class annotated by @AutoConfiguration. Plugin triggers when "gradle compileJava" is run.

Closes #722

## Summary by CodeRabbit

* **New Features**
  * Added automatic discovery of Spring Boot auto-configuration classes.
  * Spring Boot 2 and 3 starter modules now generate the required auto-configuration imports file during compilation.
  * Added support for registering the auto-configuration Gradle plugin.

* **Build Improvements**
  * Enhanced class scanning and generation of auto-configuration metadata.
  * Added Maven Central to the plugin build repositories.


mipo256 (commented):
Thanks mate @sygkanrt30 !

Before I dive into the review, the very first thing - we're writing pugins in Kotlin. If we're going to write them also in java that would cause us to create a new source set, configure this, configure that and so on.

Is there a specific reason why you've chosen to write plugin in Java?

sygkanrt30:
@mipo256 The reason is I don't know Kotlin.(((

mipo256:
There is a chance to learn it now mate @sygkanrt30 :) Do not worry that much. If you have not written any code in Kotlin still, here is your chance to learn.

> Every great developer you know got there by solving problems they were unqualified to solve until they actually did it.

:wink:

sygkanrt30:
Well, I'd love to, but it's going to take longer. @mipo256

mipo256:
Not a big deal, we're not in haste for this task @sygkanrt30

sygkanrt30:
It wasn't so difficult to rewrite Kotlin from Java.

buildSrc/src/main/kotlin/autoconfig/generator/AutoConfigFileWriter.kt
```diff
@@ -0,0 +1,64 @@
...
+    @Throws(IOException::class)
+    fun write(outputFile: File, classes: List<String>) {
+        ensureParentDirectoryExists(outputFile)
+
+        if (classes.isEmpty()) {
+            Files.deleteIfExists(outputFile.toPath())
+            return
+        }
+
+        val content = HEADER + classes.joinToString("\n") + "\n"
```
mipo256: Let's use the `File.separatorChar` here, just in case. We do not know the OS of all the contributors and the OS on CI even 😄

buildSrc/src/main/kotlin/autoconfig/generator/AutoConfigFileWriter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    @Throws(IOException::class)
+    private fun ensureParentDirectoryExists(file: File) {
+        val parent = file.parentFile
+        if (parent != null && !parent.exists()) {
+            Files.createDirectories(parent.toPath())
+        }
+    }
+
+    companion object {
```
mipo256: I know that by conventions the companion object should go as the last member, but we're making it the first for various reasons. I would stick to that.

mipo256 (changes requested):
So, we have some solid draft, but it needs improvement.

1. As noted by CodeRabbit the task is not configured anyhow. It must auto-run after the Java sources compilation. 
2. The approach that this PR takes is a bit dangerous. Parsing plaintext to construct the name of the public class - I think we can and should not do that ourselves, it is extremely error-prone. 

We use java 25 to build monorepo, so, I guess we can try to look into Class File API for parsing the `.class` and also to account for the `@AutoConfiguration`. Again, if we assume the task would run after sources compilation, we should have the access to the compiled java bytecode and use either ClassFile API or maybe some other thing like JavaAssist.

buildSrc/src/main/kotlin/autoconfig/generator/AutoConfigFileWriter.kt
sygkanrt30: @mipo256 What does it have to do with `File.separatorChar` if I need line break? As far as I know `File.separatorChar` is about path separator.

> The file should list your configuration classes, with one class name per line

This line is taken from the SB documentation.

mipo256:
We're going to get back for this PR ones 1.1 is shipped @sygkanrt30, do not worry.

sygkanrt30:
Sure,  @mipo256. And what about this one #1307?

PR #1446, opened 2026-07-23
GH-1445 Detection of blocking `KafkaTemplate` calls inside transactions for SB-3

### LINKED ISSUE #1445: [Feature] Detection of blocking `KafkaTemplate` calls inside transactions for SB-3

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Added Kafka-based monitoring for message `send` operations, including topic/target resolution and send duration tracking.
  * Kafka templates are now automatically instrumented when Kafka is available.
* **Bug Fixes**
  * Monitoring is skipped when no transaction is active and for non-send operations.
  * Prevents duplicate instrumentation; unsupported topic info results in an empty target, and `sendOffsetsToTransaction` is passed through without recording.
* **Tests**
  * Expanded Kafka monitoring, auto-configuration conditional registration, and AOP proxy/idempotency coverage.

### DIFF
--- a/sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/persistence/kafka/ExternalCallKafkaSendInterceptor.java
+++ b/sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/persistence/kafka/ExternalCallKafkaSendInterceptor.java
@@ -1,0 +1,111 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.sbs.spring.core.persistence.kafka;
+
+import java.lang.reflect.Method;
+
+import org.aopalliance.intercept.MethodInterceptor;
+import org.aopalliance.intercept.MethodInvocation;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.kafka.core.KafkaTemplate;
+import org.springframework.kafka.support.KafkaHeaders;
+import org.springframework.messaging.Message;
+
+import com.axelixlabs.axelix.common.domain.insights.TypeExternalCall;
+import com.axelixlabs.axelix.sbs.spring.core.persistence.SimpleExternalCallRecord;
+import com.axelixlabs.axelix.sbs.spring.core.persistence.transaction.TransactionAccessor;
+
+/**
+ * {@link MethodInterceptor} that records every {@code send} performed by a {@link KafkaTemplate} while a
+ * transaction is open, so that the Kafka calls made inside the transaction become visible and feed its
+ * min/max/avg statistics. It measures the {@code send()} call itself — the time it blocks the transaction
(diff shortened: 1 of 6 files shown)


sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/persistence/kafka/ExternalCallKafkaSendInterceptor.java
```diff
@@ -0,0 +1,111 @@
...
+                    new SimpleExternalCallRecord(TypeExternalCall.KAFKA, target, duration / 1_000_000));
+        }
+    }
+
+    // Matches only message sends. Excludes sendOffsetsToTransaction, which also starts with "send" but
+    // commits consumer offsets into a Kafka transaction rather than publishing a message (it has no topic).
+    private static boolean isSendMethod(Method method) {
+        String name = method.getName();
+        return name.equals("send") || name.equals("sendDefault");
+    }
```
sergeycherkasovv, line 82: In this case, we verify which method was used—specifically the [send(...) method](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L109) and the [sendDefault(...) method](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L72), which send data to a specific `topic` that we define as `target` in `SimpleExternalCallRecord`.

sbs/axelix-spring-boot-3-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/persistence/kafka/ExternalCallKafkaSendInterceptor.java
```diff
@@ -0,0 +1,111 @@
...
+                return topicName;
+            }
+            if (first instanceof ProducerRecord<?, ?> producerRecord) {
+                return producerRecord.topic();
+            }
+            if (first instanceof Message<?> message) {
+                Object topic = message.getHeaders().get(KafkaHeaders.TOPIC);
+                return topic != null ? topic.toString() : defaultTopicOrEmpty();
+            }
+        }
```
sergeycherkasovv, line 102: Since the `send(...)` method has multiple overloads, the destination can be specified in three different ways:
- the `topic` is passed as a [String ](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L109) 
- it is contained within a [ProducerRecord](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L149) 
- or it comes in the `kafka_topic` header of a [Message](https://github.com/spring-projects/spring-kafka/blob/main/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaOperations.java#L160). In the last case, the topic argument may be absent entirely—in that case, it is taken from the template’s default topic.

sergeycherkasovv:
While working on this task, I came up with a few thoughts:
  - At the moment, we track asynchronous Kafka requests, but as described in [this article](https://www.baeldung.com/spring-kafka-request-reply-synchronous), Kafka can also be used in synchronous mode. I’m not sure we need to add tracking for that behavior at this point, since it’s hard for me to imagine this being a common Kafka usage pattern.
  - We also need to take Kafka’s asynchronous nature into account: `send(...)` only buffers the message and returns a `Future`, while the broker acknowledgment arrives later on another thread. As a result, the statistics (`avg, min, max`) will often show low values

PS: once this PR is accepted, I’ll create a similar PR for Spring Boot 2.

sbs/axelix-spring-boot-3-starter/src/test/java/com/axelixlabs/axelix/sbs/spring/core/persistence/kafka/ExternalCallKafkaSendInterceptorTest.java
```diff
@@ -0,0 +1,238 @@
...
+        transactionAccessor.recordNewTransactionStarted();
+
+        // when.
+        subject.invoke(invocation);
+
+        // then. The call is passed through without being recorded.
+        verify(invocation).proceed();
+        TransactionExecutionProfile profile = transactionAccessor.recordTransactionCompletion();
+        assertThat(profile.getRecordedExternalCalls()).isEmpty();
+    }
```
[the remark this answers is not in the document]
sergeycherkasovv: No, in this case we do not need `@Nested`

Issue #1466, opened 2026-07-29
[Refactoring] Prevent user creation/rename from reusing the reserved super-admin username


NikitaKirilloff:

NikitaKirilloff:
closed as completed

PR #1467, opened 2026-07-29
GH-1466 Prevent user creation/rename from reusing the reserved super-admin username

NikitaKirilloff:


mipo256:
Cherry-picked from `master`

PR #1493, opened 2026-08-09
GH-1492 Move the built-in roles from Java constants into the database

### LINKED ISSUE #1492: [Feature] Move the `built-in` roles from Java constants into the database
**1. Create a new schema:**
- roles table
- authorities table
- roles_authorities table — maps roles to authorities
- users_roles table — maps users to roles

**2. Add the default roles** — `ADMIN`, `EDITOR`, and `VIEWER` — in a separate migration.

**Important: Do not remove the roles column**

Update UserEntity so that roles are represented through the new user-to-role mapping table instead of the existing reference. The mapping table must be populated based on the current contents of the roles column.

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Added persistent role management with role lookup and authority resolution.
  * Added built-in `VIEWER`, `EDITOR`, and `ADMIN` roles.
  * Added default authorities and role-based authority assignments.
  * Added support for role origins including built-in, web interface, and YAML.
  * Added role inheritance and user-role assignment support.

* **Database**
  * Added migrations for roles, authorities, relationships, and legacy role data across supported databases.

### DIFF
--- a/master/src/main/java/com/axelixlabs/axelix/master/domain/RoleOrigin.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/domain/RoleOrigin.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.domain;
+
+/**
+ * Origin of a role, i.e. where the role came from initially.
+ *
+ * @author Sergey Cherkasov
+ */
+public enum RoleOrigin {
+    /**
+     * Default role.
+     */
+    BUILT_IN,
+
+    /**
+     * Role created by a user through the Axelix Master UI.
+     */
+    WEB_UI,
+
+    /**
+     * Role declared in a roles YAML manifest.
+     */
+    YAML
+}
--- a/master/src/main/java/com/axelixlabs/axelix/master/domain/UserRoleName.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/domain/UserRoleName.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.domain;
+
+/**
+ * Single role assignment, used to read the assignments of many users at once without a query per user.
+ *
+ * @param userId   Identifier of the user the role is granted to.
+ * @param roleName Name of the granted role.
+ *
+ * @author Sergey Cherkasov
+ */
+public record UserRoleName(String userId, String roleName) {}
--- a/master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleService.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleService.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.service.state;
+
+import java.util.Objects;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.jspecify.annotations.NullMarked;
+import org.jspecify.annotations.Nullable;
+
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import com.axelixlabs.axelix.common.auth.core.Authority;
+import com.axelixlabs.axelix.common.auth.core.DefaultAuthority;
+import com.axelixlabs.axelix.common.auth.core.DefaultRole;
+import com.axelixlabs.axelix.common.auth.core.Role;
+import com.axelixlabs.axelix.master.domain.RoleEntity;
+import com.axelixlabs.axelix.master.repository.RoleRepository;
+
+/**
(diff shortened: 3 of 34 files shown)


master/src/main/java/com/axelixlabs/axelix/master/domain/RoleOrigin.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Origin of a role, i.e. where the role came from initially.
+ *
+ * @author Sergey Cherkasov
+ */
+public enum RoleOrigin {
+    /**
+     * Default role.
+     */
+    BUILT_IN,
```
sergeycherkasovv, line 29: Removing the default roles from `DefaultRole` will be handled in a separate PR, as it requires changes to other files and would unnecessarily increase the scope of this PR.

master/src/main/java/com/axelixlabs/axelix/master/domain/RoleOrigin.java
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Role created by a user through the Axelix Master UI.
+     */
+    WEB_UI,
+
+    /**
+     * Role declared in a roles YAML manifest.
+     */
+    YAML
```
sergeycherkasovv, line 39: I’m not sure whether we should keep these origin values in the OSS module. However, I’m also unsure about the correct approach: if OSS contains only `BUILT_IN`, while `WEB_UI` and `YAML` are defined in Enterprise, the OSS enum may fail to deserialize values written by Enterprise because it does not recognize them. On the other hand, keeping all values in OSS allows us to validate role origins))

master/src/main/java/com/axelixlabs/axelix/master/domain/UserRoleName.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Single role assignment, used to read the assignments of many users at once without a query per user.
+ *
+ * @param userId   Identifier of the user the role is granted to.
+ * @param roleName Name of the granted role.
+ *
+ * @author Sergey Cherkasov
+ */
+public record UserRoleName(String userId, String roleName) {}
```
sergeycherkasovv, line 28: We need to return a list of users together with their roles. Currently, the `users_roles` table maps users to roles by ID, but the API should return readable role names.

To avoid querying roles separately for each user, we can fetch all `user ID — role name` pairs in a single query by joining the `roles` table. `UserRoleName` was introduced as a projection for these two columns.

The implementation will be added in a separate PR. For now, I’m just sharing my thoughts in case I’m missing something. 🙂

master/src/main/java/com/axelixlabs/axelix/master/domain/RoleOrigin.java
mipo256: I think we should opt for the following:

This entire `enum` should move to the Enterprise module. The entity itself and the migrations for them should reside in the OSS (as they are). It will not be a problem to use `BUILT_IN` in the OSS migrations, since in the unified distribution that is still going to be present anyway.

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleService.java
```diff
@@ -0,0 +1,79 @@
...
+public class DatabaseRoleService implements RoleService {
+
+    private final RoleRepository roleRepository;
+
+    public DatabaseRoleService(RoleRepository roleRepository) {
+        this.roleRepository = roleRepository;
+    }
+
+    @Override
+    public Optional<Role> findByName(String name) {
```
mipo256, line 56: I'd optimize that approach. By creating a singel query with JOIN that fetches the role in a way we need. The contract would be the same though

master/src/main/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleService.java
```diff
@@ -0,0 +1,79 @@
...
+        return roleRepository.findAuthorityNamesByRoleName(roleName).stream()
+                .map(this::safeAuthorityFromString)
+                .filter(Objects::nonNull)
+                .collect(Collectors.toSet());
+    }
+
+    @Nullable
+    private Authority safeAuthorityFromString(String name) {
+        try {
+            return DefaultAuthority.valueOf(name);
```
mipo256, line 74: That is wrong. And it would not work, since we're gonna have `EnterpriseAuthority`

master/src/main/java/com/axelixlabs/axelix/master/domain/UserRoleName.java
mipo256: Adding that DTO projection in the `domain` package is just wrong. This is not a part of the `domain`. That is just a convinient data holder object. That is not something fundamental to the system, like the notion of User or Role

master/src/main/resources/db/changelog/mysql/changelog/1.1/004-create-roles-authorities-table.xml
```diff
@@ -0,0 +1,27 @@
...
+    </preConditions>
+
+    <createTable tableName="roles_authorities">
+      <column name="role_id" type="VARCHAR(36)">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="pk_roles_authorities"
+                     foreignKeyName="fk_roles_authorities_role" references="roles(id)" deleteCascade="true"/>
+      </column>
+      <column name="authority_id" type="VARCHAR(36)">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="pk_roles_authorities"
+                     foreignKeyName="fk_roles_authorities_authority" references="authorities(id)" deleteCascade="true"/>
```
mipo256, line 23: Let's create the composite primary key here in a more explicit manner, right now the way it is done is not clear. I'd create via a separate XML tag ot make it clearer for the reader.

master/src/main/resources/db/changelog/mysql/changelog/1.1/005-create-users-roles-table.xml
```diff
@@ -0,0 +1,27 @@
...
+      </not>
+    </preConditions>
+
+    <createTable tableName="users_roles">
+      <column name="user_id" type="VARCHAR(36)">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="pk_users_roles"
+                     foreignKeyName="fk_users_roles_user" references="users(id)" deleteCascade="true"/>
+      </column>
+      <column name="role_id" type="VARCHAR(36)">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="pk_users_roles"
```
mipo256, line 22: ditto: same for composite primary key

master/src/main/resources/db/changelog/mysql/changelog/1.1/006-create-roles-parents-table.xml
```diff
@@ -0,0 +1,27 @@
...
+  <changeSet id="1.1-006-mysql-create-roles-parents-table" author="Axelix Team">
+    <preConditions onFail="MARK_RAN" onError="HALT">
+      <not>
+        <tableExists tableName="roles_parents"/>
+      </not>
+    </preConditions>
+
+    <createTable tableName="roles_parents">
+      <column name="role_id" type="VARCHAR(36)">
+        <constraints nullable="false" primaryKey="true" primaryKeyName="pk_roles_parents"
```
mipo256, line 18: ditto: composite primary key

master/src/main/resources/db/changelog/mysql/changelog/1.1/007-insert-default-authorities.xml
```diff
@@ -0,0 +1,39 @@
...
+            http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-latest.xsd">
+
+  <changeSet id="1.1-007-mysql-insert-default-authorities" author="Axelix Team">
+    <preConditions onFail="MARK_RAN" onError="HALT">
+      <rowCount tableName="authorities" expectedRows="0"/>
+    </preConditions>
+
+    <insert tableName="authorities">
+      <column name="id" value="00000000-0000-0000-0000-0000000000a1"/>
+      <column name="name" value="ENV_VALUES_READ"/>
```
mipo256, line 16: for now such authorities names are fine. Although, we're going to revisit them them before 1.1.

master/src/main/resources/db/changelog/mysql/changelog/1.1/008-insert-default-roles.xml
```diff
@@ -0,0 +1,33 @@
...
+    <insert tableName="roles">
+      <column name="id" value="00000000-0000-0000-0000-0000000000b2"/>
+      <column name="name" value="EDITOR"/>
+      <column name="description" value="Performs runtime operations on the monitored applications, including destructive ones such as clearing caches."/>
+      <column name="role_origin" value="BUILT_IN"/>
+    </insert>
+    <insert tableName="roles">
+      <column name="id" value="00000000-0000-0000-0000-0000000000b3"/>
+      <column name="name" value="ADMIN"/>
+      <column name="description" value="Everything an editor can do, plus reading sensitive configuration values."/>
```
mipo256, line 29: That description is fine for now, probably is going to be revisited even before 1.1

master/src/main/resources/db/changelog/mysql/changelog/1.1/010-migrate-user-roles-to-users-roles.xml
```diff
@@ -0,0 +1,28 @@
...
+      <rowCount tableName="users_roles" expectedRows="0"/>
+    </preConditions>
+
+    <comment>
+      Fills users_roles from the legacy users.roles JSON column. The column itself is intentionally left in place:
+      it is kept up to date so that a rollback to a previous Axelix version does not lose the role assignments.
+    </comment>
+
+    <sql>
+      INSERT INTO users_roles (user_id, role_id)
```
mipo256, line 20: We need to test that migration somehow...

mipo256 (changes requested):
I'm gonna make it to master myself, thanks @sergeycherkasovv

mipo256 (approved):

Issue #1502, opened 2026-08-11
[Feature] Ability to enable structured (JSON) logging in Axelix Master


dima-bzz:
Master currently only supports Spring Boot's default plain-text console logging. There's no way to enable structured (JSON) output, which matters for containerized/production deployments where logs get shipped to a log aggregator (ELK, Graylog, Logstash, etc.) and need to be machine-parseable rather than plain text.

Master should expose a way to enable structured logging, consistent with the rest of its configuration surface under `axelix.master.*`.

dima-bzz:
@mipo256 Should this just expose a way to select among Spring Boot's existing structured formats (ecs/gelf/logstash), or is a custom Axelix format?

mipo256:
I think custom format by Axelix will be a bit of an overengineering for now. I think for now, we can go for just ECS. Other formats will be added later, on demand. WDYT @dima-bzz?

I propose this solution to not inflate the PR in size, and to not introduce features that may be needed far in the future

dima-bzz:
@mipo256 Agreed, ECS-only for now makes sense.

Implementation plan: an `EnvironmentPostProcessor` that translates a boolean `axelix.master.*` property into `logging.structured.format.console=ecs`, before the logging system initializes. `axelix.master.logging.environment`, when set, is forwarded to the native `logging.structured.ecs.service.environment` field.

For the property name itself, e.g. `axelix.master.logging.json=true` or `axelix.master.logging.structured=true` - a plain boolean rather than a format-selector string, since there's only one format to turn on right now; if/when format choice is added later, that's a new property, not a migration of this one.

PR #1504, opened 2026-08-12
Add Spring Cloud Config Client support for Master via axelix.master.config-server properties.

### LINKED ISSUE #1491: [Feature] Provide the integration with Spring Cloud Config in Axelix Master
Similar to #1489, we also need to be able to take properties from Spring Cloud Config in Axelix Master.

sygkanrt30:
Wire Spring Cloud Config Client so operators can point Master at an external Config Server using axelix.master.config-server.*. Values are bridged to spring.cloud.config.* in application configuration. Active profiles continue to come from the Spring Env.

Closes #1491

## Summary by CodeRabbit

* **New Features**
  * Added support for centralized cloud configuration.
  * Added configurable connection settings and optional imports for configuration services.
  * Added automatic configuration activation based on deployment settings.
  * Added fallback defaults to support reliable startup when centralized configuration is unavailable.
* **Configuration**
  * Disabled centralized cloud configuration for local development.
  * Added application-level defaults for flexible deployment environments.
* **Bug Fixes**
  * Improved handling of configuration options regardless of letter casing.


master/src/main/resources/application-cloudconfig.yaml
mipo256, line 1: What is the reason the configuration for Spring Cloud Config resides in the special `application-cloudconfig.yaml` file? It is going to be activated when the `cloudconfig` profile is enabled, but as of now we do not have any profiles for Axelix Master.

I guess the rationale was (and correct me if I am wrong) to just offload the configuration related to Spring Cloud Config to a separate YAML configuration file, so that the main `application.yaml` is not polluted with the properties that it may not need, is that right?

If it is, then I understand that rationale, and I think this is a good thought direction, but I think that this needs to be addressed in the separate task, and it needs to be designed separately. It is unclear to me that relying on Spring profiles is the best choice here. 

**Bottom Line:** I'd merge this configuration into the main `application.yaml` file. WDYT @sygkanrt30 ? That would also automatically resolve the propeties that are re-declared in this file, such as `server.port`, and `spring.application.name`

master/src/main/resources/application.yaml
```diff
@@ -7,6 +7,8 @@ spring:
     compatibility-verifier:
       # TODO: This is a temporary fix, until we wait for the official cloud support of boot 4.1
       compatible-boot-versions: 4.1.x
+    config:
```
mipo256: That property would go away

master/src/main/resources/application-cloudconfig.yaml
```diff
@@ -0,0 +1,28 @@
...
+      fail-fast: ${axelix.master.config-server.fail-fast}
+  config:
+    import: 'configserver:'
+  application:
+    name: axelix
+
+axelix:
+  master:
+    config-server:
+      enabled: true
```
mipo256: By default, it should be `false`. Since this config is potentially going to be merged into the `applicaiton.yaml`, this should be false.

mipo256 (changes requested):
I think we're generally headed in the good direction. Still, a couple of things to discuss @sygkanrt30

master/src/main/resources/application-cloudconfig.yaml
sygkanrt30: > I guess the rationale was (and correct me if I am wrong) to just offload the configuration related to Spring Cloud Config to a separate YAML configuration file, so that the main application.yaml is not polluted with the properties that it may not need, is that right?

Yes @mipo256, you're right! That's exactly what I was thinking.

> If it is, then I understand that rationale, and I think this is a good thought direction, but I think that this needs to be addressed in the separate task, and it needs to be designed separately. It is unclear to me that relying on Spring profiles is the best choice here.

Okay, perhaps it's more correct than to implement it in this issue.

> I'd merge this configuration into the main application.yaml file. WDYT @sygkanrt30 ?

Sure, I'll do it.

master/src/main/resources/application-cloudconfig.yaml
sygkanrt30: Of course, it's absolutely necessary now.

mipo256:
Okay, so, I've spent some on this PR and I think I have a clear direction on where we're going to take it.

First of all, I think that we should define the properties that are going to control the behavior of Axelix Master when connecting to the Spring Cloud Config server. And these options are:

`axelix.master.external-config.option` which is going to signal the option of the external configuration. For now, the only value of that flag should be `spring-cloud-config`. In the future, considering the possible upcoming addition of Hashicorp Vault (@reptiloyds Ilya, please, take note here), we're going to be able to expand the possible set of values used and add `hashicorp-vault` as well. 

If we have `axelix.master.external-config.option=spring-cloud-config`, we should just set the `spring.config.import` property value to `:configserver`, and (just in case) set `spring.cloud.config.enabled=true` (it is true by default, but I think we should set it explicitly to true just in case, to avoid future pitfalls if Cloud team will change the default on the next minor release).

If `axelix.master.external-config.option=spring-cloud-config` is not specified, then we should just not do anything, not even tackle the `spring.cloud.config.enabled`, since this property by itself has no meaning unless `spring.config.import` is specified correctly. So it's pretty simple.

Now, the thing that is important here is this - **it is almost certainly going to be quite challenging to express in the yaml configuration** (i.e. using placeholders like `${...}` and SPeL `#{...}`), and I do not think that this is the way to go.

We recently merged the PR by @dima-bzz #1502, where we concluded that such complex scenarios probably are better to be solved by the custom `EnvironmentPostProcessor`. So, you can inspire from the Dima's PR. I think adding `ExternalConfigurationEnvironmentPostProcessor` and then implementing the value specification logic in there is the right way to go.

Apart from that, all the other properties (like URI, password, username, label and so on) - I think that they should go under `axelix.master.external-config.spring-cloud-config.*` prefix. Also, I am not sure that we need to add `fail-fast` option right now. It is going to complicate the implementation a bit, and it is unclear whether we need it right away. So, my verdict is this - we may easily avoid `fail-fast` right now. If you @sygkanrt30 think that we may cheaply add it - I think we may give it a try. If not - no reason to bother here.

sygkanrt30:
Fine @mipo256 , I'll implement it this way.

mipo256:
I see you've been busy, please, ping me once this one is ready for review @sygkanrt30

sygkanrt30:
Hi @mipo256, this one is ready for review!

mipo256:
Great, thanks @sygkanrt30, I will take a look on this weekends

mipo256 (approved):
Thank you, @sygkanrt30, I have gave it a look. Here is what I think.

I perfectly understand why this dance with pre-loading the YAML was made, and I understand the problem behind it. However, it is not the approach we can take for a number of reasons:

1. We claim ([in the docs](https://axelix.io/docs/setting-up-master-ui/configuring-master/#supplying-properties)) that the Axelix Master is capable of accepting properties in various sources (env vars, `-D` vm options and so on). So, assuming that the proeprties will be in the YAML that is pre-bunded in the Fat Master JAR is just not the right way to go - it is an incorrect assumption.

2. Since we're capable of loading configs from various sources, we must support relaxed binding, and we do (again, [we state that in the docs](https://axelix.io/docs/setting-up-master-ui/configuring-master/#supplying-properties)). Currently, even if we inspect all the sources ourselves, which is not a good idea btw, wihtout using `Binder` and Boot's machinery we cannot adhere the relaxed binding, which is also a problem.

So, I have given it a couple of tries, and I think I managed to make it work in a simple fashion in the YAML (take a look at my commit in your PR).

As a result, I have decided to proceed with the simplified version, where necessary configuration is applied as the separate YAML configuration block only. We're going to add the E2E nightly test for it a bit later (I'll create a task for it, this is going to be not that easy).

**I have preserved your authorship of the commit to honor your work and time you put in.**

Thank you @sygkanrt30.

mipo256:
CC: @dima-bzz

You may need to take a look at [this comment above](https://github.com/axelixlabs/axelix/pull/1504#pullrequestreview-5028247829)

Issue #1497, opened 2026-08-10
[Feature] Ensure the correct resources cleanup in the Scheduled tasks domain in starters


mipo256:
While revieing the #1485, it became clear that we lack the appropriate resources for closure in Scheduled tasks domain in the starters.

In particular, our `ScheduledTaskService`, `ManagedScheduledTask`, `ScheduledTasksRegistry` and so on - they all probably need to implement the `Closeable` interface, since theu either:

1. Have something to close by themselves, like `ManagedScheduledTask` (has ), or `ScheduledTaskService` (has `ThreadPoolTaskExecutor`)
2. Or they have something isnide of them to close, like `ScheduledTasksRegistry` does

In either case, we need to ensure the appropriate closure of resources on application cotnext shutdown

sygkanrt30:
Hey @mipo256, I'd like to do that! 

So, you can safely assign this ticket to me.

mipo256:
Sure @sygkanrt30

sygkanrt30:
@mipo256 I'm going to create 3 sepatate PR for sb 2,3,4. 

If the changes fit in 500 lines, should I create only one PR?

mipo256:
I think three separate PRs are quite fine. The smaller the PR the generally better :) @sygkanrt30 

Apart from that, if we're going to pinpoint something important, we're presumably going to spend less time re-implementing the solution.

PR #1508, opened 2026-08-12
Implement AutoCloseable interface for graceful resource cleanup in Scheduled tasks domain for SB 3

sygkanrt30:
Introduce automatic management of scheduled tasks lifecycles by leveraging Spring's default shutdown hooks. This ensures all active scheduled futures are explicitly cancelled and memory-managed registries are cleared upon application context termination.

Integration tests using ApplicationContextRunner and DirtiesContext have been added to verify that task states transition cleanly during the shutdown phase.

Related tickets #1497

## Summary by CodeRabbit

* **New Features**
  * Scheduled tasks now shut down cleanly when the application context or task service closes.
  * Managed tasks are cancelled and removed from the registry during shutdown.
  * Shutdown continues safely even if an individual task fails to close.

* **Bug Fixes**
  * Prevented scheduled tasks from remaining active after application shutdown.

* **Tests**
  * Added coverage validating task cancellation, shutdown behavior, and registry cleanup.


mipo256 (approved):
That's good. I'm a bit troubled about the `@DirtiesContext` usage, but I do not think that there is any other way out that we would like. 

We may try to cherry-pcik and create the context we want for tests, like the one with the `ApplicationContextRunner` or something, although, it probably will not be better and it will introduce even more code, and will not really result in much perf. improvement. 

So, LGTM.

mipo256:
Thanks, @sygkanrt30 !

That has been merged.

PR #1510, opened 2026-08-14
Add structured ECS logging configuration for Master.

### LINKED ISSUE #1502: [Feature] Ability to enable structured (JSON) logging in Axelix Master

dima-bzz:
Introduce axelix.master.logging.json, which enables Spring Boot's ECS structured console logging via an EnvironmentPostProcessor, and axelix.master.environment, forwarded into it when logging.json is enabled, and document both in the Master configuration reference.

Closes #1502

## Summary by CodeRabbit

* **New Features**
  * Added optional ECS-compatible JSON logging for Master, disabled by default.
  * Added configuration for structured logging and deployment environment metadata.
  * Environment information can be included in structured log output when configured.
  * Plain-text logging remains available when structured logging is disabled.

* **Documentation**
  * Added English and Russian documentation covering logging options, defaults, and configuration examples.
  * Added notices for upcoming releases in relevant configuration sections.

* **Bug Fixes**
  * Preserved explicitly configured logging properties when structured logging is enabled.


docs/docs/setting-up-master-ui/configuring-master/configuring-master.mdx
```diff
@@ -32,6 +32,16 @@ the JWT signing key has no default at all. Override both — together with `axel
 over HTTPS — before exposing Master to anyone other than yourself.
 :::
 
+### General
+
+<div className={styles.ConfigTable}>
+
+| Property                    | Default | Description                                                                             |
+|-------------------------------|---------|--------------------------------------------------------------------------------------------|
+| `axelix.master.environment` | empty   | Name of the environment this Master instance runs in, for example `production` or `staging`. |
```
mipo256: I think I understand the rationale behind that move, i.e. the rationale on why we would want to introduce this as a separate, non-logging-related property. 

I think I generally agree with the choice of it, although we're risking a bit to fall into the premature optimization territory, but all things considered - let's put it in there.

The only thing I would do as well is that I would give a reference to the "Structed Logging" section in the description. The reason for that is that right now, it is very unclear for the external reader:

1. Why does he need to set this property? 
2. What does it affect? 
3. What would happen if he/she would set it wrong? And so on

We need to remove those questions/concerns by clearly stating in the description of the property that for now, it is ONLY used for that single purpose. If any other use comes up in the future (which is highly probable) - we're going update the description, but for now - let's clarify it in the docs.

The same applies for the i18n-ized docs :)

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/logging/AxelixStructuredLoggingEnvironmentPostProcessor.java
mipo256, line 1: I'm not exactly sure that we need the `EnvironmentPostProcessor` here. Can we just get away with simply mapping the properties? Like, we already do the direct forwarding in the `application.yaml`. If we can, it would be:

1. Aligned with the current technique used to map properties from Axelix to Spring Boot ones.
2. And it would also be simpler (just less code and less moving parts in the master).

I mean, right now, I do not really see much need for the custom `EnvironmentPostProcessor`. Maybe I am wrong, tell me if I am @dima-bzz. Can we not introduce it _yet_?

docs/docs/setting-up-master-ui/configuring-master/configuring-master.mdx
```diff
@@ -190,6 +200,21 @@ properties that go *into* the heartbeat (`master-url`, `instance-actuator-url`,
 
 </div>
 
+### Structured logging
+
+By default, Master logs plain text to the console. Set `axelix.master.logging.json=true` to switch to
```
mipo256: I would clarify on what we mean by the _"console"_. Same for the i18n-ized docs :)

```suggestion
By default, Master logs plain text to the console (i.e. the process stdout). Set `axelix.master.logging.json=true` to switch to
```

docs/docs/setting-up-master-ui/configuring-master/configuring-master.mdx
```diff
@@ -190,6 +200,21 @@ properties that go *into* the heartbeat (`master-url`, `instance-actuator-url`,
...
+[Elastic Common Schema](https://www.elastic.co/guide/en/ecs/8.11/ecs-reference.html) (ECS) structured JSON logging
+instead — the only structured format Master supports today. ECS is a JSON-based logging format; reach for it when logs
+are shipped to an aggregator such as Elasticsearch or a similar ECS-aware pipeline.
+
+<div className={styles.ConfigTable}>
+
+| Property                     | Default | Description                                                               |
+|-------------------------------|---------|------------------------------------------------------------------------------|
+| `axelix.master.logging.json` | `false` | Switch console logging from plain text to ECS-formatted structured JSON.     |
```
mipo256: Lets name the property `axelix.master.logging.json.enabled`. A couple of reasons for that:

1. If we later would introduce (and it probably would happen) `axelix.master.logging.json.format`, then having `axelix.master.logging.json: true` in the YAML will cause us troubles.

2. And apart from that, there is a general naming convention (in Spring, and in Axelix we try to follow), that properties that turn something on/off end with `enabled` postfix.

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/logging/AxelixStructuredLoggingIntegrationTest.java
mipo256, line 1: This test case is generally designed pretty well @dima-bzz. 

Still, I think that the way assertions are done in this test are a bit off, since the test does not in fact check that we output a valid JSON, so it can be a bit deceptive. I think we can use JsonUnit here, as we already do in master as well. We can omit the values of the fields that we do not care about, and only check the ones that we can actaully durably check, and the fact that the result is in fact a JSON.

WDYT?

mipo256 (changes requested):
I think we're headed in the right direction thank you @dima-bzz.

Still, there are some things to work on. Please, read my comments

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/logging/AxelixStructuredLoggingEnvironmentPostProcessor.java
dima-bzz: @mipo256 I tried plain property mapping first, it doesn't work.

`logging.structured.format.console` is not a flag — it's a format id (`ecs`, `gelf`, `logstash`, or a `StructuredLogFormatter` FQCN), and there is no value meaning "off". So `logging.structured.format.console: ${axelix.master.logging.json.enabled}` fails: `false` goes into the formatter factory as an unknown format. Spring Boot has no way to disable structured logging yet — spring-projects/spring-boot#45407, open, `pending-design-work`, milestone `4.x`. There's also no `spring.config.activate.on-property` (only `on-profile` / `on-cloud-platform`), so a conditional config document isn't an option either.

The only pure-YAML workaround is nested-placeholder indirection:

```yaml
axelix:
  master:
    logging:
      json:
        enabled: false
  internal:
    console-format-true: ecs
    console-format-false: ""

logging.structured.format.console: ${axelix.internal.console-format-${axelix.master.logging.json.enabled:false}:}
```

It resolves, but I don't want to ship it:

1. No relaxed binding — the key is string-concatenated, so `TRUE` or `True` builds `console-format-TRUE`, falls through to the empty default and silently leaves logging in plain text.
2. It relies on "empty format == plain text", which is an implementation detail of how Boot picks between `console-appender.xml` and `structured-console-appender.xml`, not a contract.
3. `axelix.internal.console-format-*` become visible, overridable properties.

The one option that does remove the post-processor is dropping the boolean for a format string:

```yaml
axelix.master.logging.json.format:   # empty = plain text
logging.structured.format.console: ${axelix.master.logging.json.format:}
```

So: (a) keep the post-processor, (b) switch to the format-string property and drop it, or (c) keep the boolean with the placeholder trick? I lean to (a).

Agreed on `axelix.master.logging.json.enabled` — will rename in the next push.

docs/docs/setting-up-master-ui/configuring-master/configuring-master.mdx
dima-bzz: I will update the documentation

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/logging/AxelixStructuredLoggingEnvironmentPostProcessor.java
mipo256: I see. Alright, I agree. I think that for now we indeed need this `EnvironmentPostProcessor`. We actually can do the dance with proeprties simpler than you described, but in any case, that would require using SPeL, and that is essentially logic inside the confiugration and that is generally a pretty sold path to hell in the long run.

So, let's leave it. I agree.

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/logging/AxelixStructuredLoggingEnvironmentPostProcessorTest.java
```diff
@@ -0,0 +1,151 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Unit tests for {@link AxelixStructuredLoggingEnvironmentPostProcessor}.
+ *
+ * @author Dmitry Mazurov
+ */
+class AxelixStructuredLoggingEnvironmentPostProcessorTest {
+
+    private static final String STRUCTURED_FORMAT_CONSOLE_PROPERTY = "logging.structured.format.console";
```
mipo256: I think we should re-use the constants in the `EnvironmentPostProcessor`. I'll make them public and then re-use them here

docs/i18n/ru/docusaurus-plugin-content-docs/current/setting-up-master-ui/configuring-master/configuring-master.mdx
```diff
@@ -36,6 +36,16 @@ import styles from './styles.module.css';
 себя.
 :::
 
+### Общие настройки
+
+<div className={styles.ConfigTable}>
+
+| Свойство                    | По умолчанию | Описание                                                                                       |
```
mipo256, line 46: Thats good, the only thing that's left here is to add the `UpcomingReleaseNotice` to the sections. But apart from that its good.

mipo256 (approved):
Okay, apart fro mminor polishing, we're good

mipo256:
@cla-assistant recheck

mipo256:
recheck

mipo256:
Alright, for some reason CLA Assistant does not want to check wether the CLA being signed, still I know that it was (reference PR: https://github.com/axelixlabs/axelix/pull/1494). I'm gonna merge it.

Thanks, @dima-bzz !

PR #1518, opened 2026-08-16
Add configurable Prometheus common tags for Master.

### LINKED ISSUE #1495: [Feature] Expose /actuator/prometheus endpoint on Master
Master currently exposes only the `health` actuator endpoint (`management.endpoints.web.exposure.include: [ "health" ]` in `master/src/main/resources/application.yaml`). Add support for                          
  `/api/actuator/prometheus` so Master's own metrics can be scraped by Prometheus.

dima-bzz:
Master now exposes a Prometheus scrape endpoint at /api/actuator/prometheus, gated by axelix.master.monitoring.prometheus.enabled (default false). Common tags can be attached to every exported metric via axelix.master.monitoring.prometheus.tags.<name>=<value>, wired into Micrometer through a MeterRegistryCustomizer bean.

Closes #1495

## Summary by CodeRabbit

* **New Features**
  * Added optional Prometheus metrics support through the `/api/actuator/prometheus` endpoint.
  * Added configurable common tags for published metrics.
  * Added JVM metrics and region tag visibility when monitoring is enabled.

* **Documentation**
  * Documented configuration, defaults, endpoint access, security considerations, and YAML examples in English and Russian.

* **Bug Fixes**
  * Prometheus monitoring is disabled by default and returns an appropriate not-found response when unavailable.


dima-bzz:
Should the prefix be `axelix.master.management` instead of `axelix.master.monitoring`? WDYT @mipo256

mipo256:
> Should the prefix be `axelix.master.management` instead of `axelix.master.monitoring`? 

I actually think `axelix.master.monitoring` is quite fine, TBH. Let's leave it as it is @dima-bzz

dima-bzz:
If we add something else later — a separate actuator port, for example — `axelix.master.management.port` fits, while `axelix.master.monitoring.port` doesn't really.

mipo256:
> If we add something else later — a separate actuator port, for example — `axelix.master.management.port` fits, while `axelix.master.monitoring.port` doesn't really.

I do not really think that this would make much difference. There is no clear objective way of telling what is more intuitive. So I would stick with `monitroing` and just keep it this way. At the end of the day, this is probably an inconsequential choice @dima-bzz

dima-bzz:
I added this warning to the docs:

> Because the endpoint is unauthenticated, anyone who can reach Master over the network can read metric values and any configured common tags. Restrict `/api/actuator/prometheus` to trusted Prometheus infrastructure with a network policy, ingress rule, or an equivalent perimeter control.

While writing it I thought a simpler way to get that protection is to move the actuator to a separate port (`management.server.port`). Then it's "don't publish port 8081" instead of a path rule that has to be repeated in every ingress and proxy. Out of scope here, but maybe worth a separate issue. WDYT @mipo256

mipo256:
> While writing it I thought a simpler way to get that protection is to move the actuator to a separate port (management.server.port). Then it's "don't publish port 8081" instead of a path rule that has to be repeated in every ingress and proxy. Out of scope here, but maybe worth a separate issue. WDYT @mipo256

I think that your suggestion is a good one, and it would be helpful in real production environments. Isolation of the port is generally simpler than relying on some path-based networking policies, etc.
 
And I agree that this should be handled by a separate task. Please, ping me, once the PR is ready for review @dima-bzz

dima-bzz:
@mipo256 ready for review

master/build.gradle.kts
```diff
@@ -98,6 +99,7 @@ dependencies {
     testFixturesApi("digital.pragmatech.testing:spring-test-profiler:0.1.2")
     testFixturesApi("org.instancio:instancio-core:${instancioVersion}")
     testFixturesApi("net.javacrumbs.json-unit:json-unit-assertj:${jsonUnitAssertJVersion}")
+    testFixturesApi("io.prometheus:prometheus-metrics-exposition-formats")
```
mipo256: Do we really need this dependency for the `testFixtures`? I am not really sure why would we...

master/src/main/resources/application.yaml
```diff
@@ -32,14 +32,22 @@ management:
     web:
       base-path: /api/actuator
       exposure:
-        include: [ "health" ]
+        include: [ "health", "prometheus"]
```
mipo256: a small nit:

```suggestion
        include: [ "health", "prometheus" ]
```

master/src/main/resources/application.yaml
```diff
@@ -32,14 +32,22 @@ management:
...
+  prometheus:
+    metrics:
+      export:
+        enabled: ${axelix.master.metrics.prometheus.enabled:false}
```
mipo256: I think we do not need the default value at the injection point since we already specify the default value at the declaration site: 

```suggestion
        enabled: ${axelix.master.metrics.prometheus.enabled}
```

master/src/testFixtures/resources/application.yaml
```diff
@@ -38,10 +38,17 @@ management:
     web:
       base-path: /api/actuator
       exposure:
-        include: [ "health" ]
+        include: [ "health", "prometheus" ]
+  prometheus:
+    metrics:
+      export:
+        enabled: ${axelix.master.metrics.prometheus.enabled:false}
```
mipo256: Same here :)

```suggestion
        enabled: ${axelix.master.metrics.prometheus.enabled}
```

mipo256 (commented):
We're almost there. A few nitpicks, and apart from that we're pretty good.

mipo256:
That looks good now, thanks @dima-bzz ! 

This feature will be shipped in the upcoming minor release.

PR #1523, opened 2026-08-17
GH-1511: Added troubleshooting page in documentation

### LINKED ISSUE #1511: [Docs] Introduce the Troubleshooting
As it turned out, there are a couple of problems that people encounter during their Axelix installation. We need to probably introduce a separate section/page called "Troubleshooting", where we outline them. 

In particular, we would want to outline the condition (quite common one), when the end-user's actuator endpoints already have some security mechanism baked into them (Basic, OIDC or etc), and as a result, custom actuator endpoints created by Axelix may also accidentally get included in this security mechanism, which is wrong.

Axelix has its own RBAC, and Axelix endpoints (at least of of now) are not supposed to be invoked directly, rather, they are supposed to be invoked only by Axelix Master. 

Thus, the viable solution that people commonly do is that they exclude Axelix acutator endpoints from common RBAC checks, and rely purely on Axelix RBAC.

DNekhoroshev:
## Summary by CodeRabbit

* **Documentation**
  * Added a troubleshooting guide for 401 errors when accessing Axelix endpoints.
  * Explained bearer-token validation and interactions with application security filters.
  * Documented configurable actuator endpoint prefixes and a Spring Security workaround for excluding Axelix endpoints from application authentication checks.
  * Added a Russian translation of the troubleshooting guide.
  * Added the guide to the More documentation navigation.


mipo256 (approved):
I've polished the docs a bit to generalize the problem a bit. Apart from that, we're pretty good

PR #1530, opened 2026-08-19
GH-1489: Add external config support via EnvironmentPostProcessor

### LINKED ISSUE #1489: [Feature] Take secrets from vault in master
For Axelix Master, we need to take secrets from Vault

reptiloyds:
Part of #1489 .

## Summary
Introduces a mechanism to dynamically configure `spring.config.import` for external configuration sources (based on discussion #1504).

The configuration is implemented using `ExternalConfigurationEnvironmentPostProcessor`, which dynamically adds the required `spring.config.import` entries based on the available external configuration options.

The solution is also designed to be easily extended to address issue #1491: support for additional external configuration sources can be added by introducing a new value to the `ExternalConfigOption` enum.

## Notes
- Options must be provided via environment variables, not from `application.yaml`, since the processor runs before `ConfigDataEnvironmentPostProcessor`.
Running the processor after `ConfigDataEnvironmentPostProcessor` would not make sense either, as both `spring.config.import` and `application.yaml` are processed by `ConfigDataEnvironmentPostProcessor` itself.
- Had to update one of the `UserApiTest` tests because `spring-cloud-starter-vault-config` introduced Apache HttpClient, which caused `TestRestTemplate` to automatically persist cookies, AFAIU.

**P.S.** I'm not completely sure about the current approach, so I'd like to discuss it in the related issue.

## Summary by CodeRabbit

* **New Features**
  * Added optional Spring Cloud Vault integration for external configuration.
  * Vault support is disabled by default and can be enabled and configured when needed.

* **Bug Fixes**
  * Improved logout testing for requests made without authentication cookies.

* **Chores**
  * Updated local and test configurations to keep Vault disabled by default.
  * Removed an obsolete IDE Kotlin configuration file.


mipo256:
Hey @reptiloyds! We're going to check this PR shortly. Thanks!

dima-bzz:
@reptiloyds  A couple of directions worth checking:                                                                                                                                                              
                                                                                                                                                                                                                     
  1. VaultProperties remapping: instead of listing fields one by one in application.yaml, look at VaultConfigDataLocationResolver#registerVaultProperties — it uses bootstrapContext.registerIfAbsent(VaultProperties.class, ...). Pre-registering your own instance (bound from axelix.master.external-config.vault.*) via an earlier BootstrapRegistryInitializer/EnvironmentPostProcessor wins that registerIfAbsent race and covers all fields at once, no per-field mapping needed.

```java
 bootstrapContext.registerIfAbsent(VaultProperties.class,                                                                                                                                                   
                  ctx -> Binder.get(environment).bindOrCreate("axelix.master.external-config.vault", VaultProperties.class)); 
```
                                               
  2. KV backend: worth deciding between two options — (a) drop the spring.cloud.vault.kv.* convention and use explicit vault://<path> imports instead, or (b) keep the convention but populate spring.cloud.vault.kv.* by copying it from axelix.master.external-config.vault.kv.*. Note VaultKeyValueBackendProperties is always bound fresh from the hardcoded spring.cloud.vault.kv prefix (no registerIfAbsent/bootstrap-registry hook like VaultProperties has), so (b) only works via property copying, not object substitution.

reptiloyds:
@dima-bzz Thanks for the improvement suggestion!
I agree that getting rid of the manual property mapping is necessary.

I’ve almost finished implementing this improvement, but I’m concerned that the current PR is already getting too large. In the latest commit, I only removed the manual property mapping.

Would it be possible to accept the PR in its current form? I’ll open a separate PR with the remaining improvements you suggested.

dima-bzz:
Hi @reptiloyds. Take a look at #1504 and @mipo256’s comment there — same EnvironmentPostProcessor approach was tried and ended up being replaced with a YAML-based one. Might be relevant here.

dima-bzz:
A few thoughts on the implementation.

**We only need properties at startup.** Master reads secrets once, before the context comes up — nothing in the codebase uses VaultTemplate afterwards. That makes VaultAutoConfiguration dead weight: under spring.config.import=vault:// the infrastructure beans are built by VaultConfigDataLoader anyway, and the autoconfiguration only matters on the non-ConfigData path. Its @EnableConfigurationProperties(VaultProperties.class) also leaves a second VaultProperties bean around. I’d add it to spring.autoconfigure.exclude.

VaultHealthIndicatorAutoConfiguration **needs a separate decision.** The exclude above doesn’t touch it. It’s conditioned on spring.cloud.vault.enabled (matchIfMissing = true), not on our flag — with Vault off in local/test it still activates and reports DOWN forever, which on Kubernetes means restart loops. Either exclude it too, or drive management.health.vault.enabled from our own flag. Given we don’t keep a live Vault client, an indicator that pings Vault forever doesn’t buy us much.

**On the env-vars-only limitation.** That goes away with a ConfigDataLocationResolver: inside resolveProfileSpecific you get context.getBinder(), which sees contributors already processed — including the application.yaml document declaring the import. Our own VaultProperties subclass can then be registered into the bootstrap context under our prefix, and the standard loader runs unchanged. The one thing that won’t move across is spring.cloud.vault.kv.* — the backend descriptors are read straight from the binder with no override hook, so those keys have to be copied or pinned separately.

@reptiloyds happy to go through any of this in more detail — just ping me.

reptiloyds:
@dima-bzz Thanks for the clarification!

I have a question regarding `ConfigDataLocationResolver`. Am I correct in understanding that the only option is to extend `VaultConfigDataLocationResolver` and give our implementation a higher priority so that it takes precedence over the original implementation? AFAIK, only one `ConfigDataLocationResolver` can handle each `spring.config.import` location, and the others are skipped once a matching resolver is found.

dima-bzz:
@reptiloyds yes, that’s correct. ConfigDataLocationResolvers iterates the resolvers and returns on the first one whose isResolvable returns true — the others are skipped. Since VaultConfigDataLocationResolver declares no order, it’s at lowest precedence, so ours only needs an explicit one to win.

dima-bzz:
Hi @reptiloyds!

One thing worth thinking about: the lease and session lifecycle. Secrets imported through config-data are read once into the Environment at startup, so a renewed lease doesn’t change anything for the running application without refresh scope — what remains is a background renewal thread and periodic traffic to Vault. The session lifecycle is the same question: token renewal only matters if something keeps talking to Vault after startup.

So maybe both should be turned off, either in the config or in AxelixVaultConfigDataLocationResolver — the resolver would make the behaviour independent of what the user configures. Worth considering which of the two fits better.

WDYT @mipo256

master/src/main/resources/application.yaml
```diff
@@ -50,6 +56,17 @@ axelix:
         name:
         username:
         password:
+      # These properties are automatically bound to VaultProperties
+      # (see AxelixVaultConfigDataLocationResolver), except for the kv properties.
```
mipo256: This comment is good, but it does not outline the whole picture.

I'd enhance the comment here by outlining that normally, we enumerate all the properties available in Axelix with their default values in here, but since for hashicorp vault there're a lot of properties used to connect to it, we do not enumerate them theere with their defaults. All the properties that Axelix offers in order to connect to the hashicorp vault can be found in here (for `kv` vault engine), and in the `VaultProperties` class in the spring-cloud-vault

master/src/main/resources/application.yaml
```diff
@@ -4,11 +4,13 @@ server:
 
 spring:
   autoconfigure:
-    # OTLP and Prometheus metrics export are owned by OtlpTransportMetricsAutoConfiguration and
-    # PrometheusMetricsAutoConfiguration respectively, not Spring Boot's own auto-configurations.
     exclude:
```
mipo256: We no longer do exlcudes like this, see `@AxelixMasterApplication` anntotaion in master

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/externalconfig/AxelixVaultProperties.java
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.autoconfiguration.externalconfig;
+
+import org.springframework.cloud.vault.config.VaultProperties;
+
+public class AxelixVaultProperties extends VaultProperties {
```
mipo256, line 28: 1. Add the javadoc honoring the author, please
2. Apart from that, we need to clarify on why we choose that approach (in the comment I guess). Here is how I look at this properties class:

So, we typically try to make the properties that Axelix provides for the user very explicit and map them into Spring boot's native properties since **boot's contract != Axelix public contract**. Here, although we provide our own properties (see `AxelixVaultConfigDataLocationResolver`) with our own prefix, they are tightly coupled with spring cloud's properties. This is a deliberate tradeoff since if we would've create our own proeprties class and map that all up explciitely, there would be much-much more code. In our case, the amount of code is a bit less, but the overall design is a bit more shaky - **that is okay, we take taht risk**. 

im talking about the risk that spirng cloud will change the proeprty underneah and break axelix with it. I hope we're going to catch it with tests, but at very least spring cloud vault as the project is not that actively devleoped right now, so I expect not that much of the breaking changes here.

master/src/testFixtures/resources/application.yaml
```diff
@@ -1,29 +1,26 @@
 spring:
   autoconfigure:
-    # OTLP and Prometheus metrics export are owned by OtlpTransportMetricsAutoConfiguration and
-    # PrometheusMetricsAutoConfiguration respectively, not Spring Boot's own auto-configurations.
     exclude:
```
mipo256: exlucsions - ditto

mipo256 (commented):
Overall I think the design is good. Thanks @reptiloyds, @dima-bzz. The core team is going to pick it up.

Nikita, @NikitaKirilloff, please, make it to the master trunk. I have left my overall thoughts in individual comments. 

Apart from all the comments I left, I think that tests in this PR are a bit off. Please, revisit them entirely. I would've done them in similar fashion as we did with Spring Cloud Config Server, i.e. with the test-contaienr. Fortunately, in this case we have the [dedicated solution](https://java.testcontainers.org/modules/vault/) for that.

In any case, I think that testing interaction and the fact that credentials have ended up in the master's environment and are honor (e.g. mcp server is in fact enabled or something) is sufficient for now. For auth lests use the appRole since this is the most popular auth method for vault I guess.

master/src/test/java/com/axelixlabs/axelix/master/api/external/endpoint/UserApiTest.java
```diff
@@ -277,8 +278,9 @@ void shouldReturn401OnLogoutWithoutCookie() {
         HttpEntity<Void> logoutEntity = new HttpEntity<>(headers);
 
         // when.
-        ResponseEntity<String> logoutResponse =
-                restTemplate.exchange("/api/external/users/logout", HttpMethod.POST, logoutEntity, String.class);
+        ResponseEntity<String> logoutResponse = restTemplateBuilder
+                .withoutToken()
+                .exchange("/api/external/users/logout", HttpMethod.POST, logoutEntity, String.class);
```
NikitaKirilloff: I’m not sure why these changes were needed—everything works perfectly fine both locally and in CI for me. If you’re having issues with this test, please let me know. @reptiloyds

reptiloyds: Everything works fine for me now too. Not entirely sure what changed, but I think I was hitting test pollution from RestTemplate keeping cookie state across tests.

mipo256 (approved):

PR #1536, opened 2026-08-22
Ability to configure a separate port for the Prometheus endpoint.

### LINKED ISSUE #1520: [Feature] Ability to configure a separate management (actuator) port in Axelix Master
Master's actuator endpoints (`/api/actuator/health`, /`api/actuator/prometheus`) always run on the same port as the main application (server.port). There's no way to bind them to a separate port, which matters for the unauthenticated Prometheus scrape endpoint (`axelix.master.metrics.prometheus.enabled`) — operators can't isolate it to an internal-only interface without also restricting the whole application port.

dima-bzz:
`axelix.master.metrics.prometheus.port `now controls where Prometheus metrics are exposed. Left unset, it defaults to server.port and metrics are served through the actuator. Set to a different value, Prometheus starts its own dedicated HTTP server on that port instead; 0 picks a random free port for that server.

Closes #1520

## Summary by CodeRabbit

* **New Features**
  * Prometheus metrics can run on a dedicated HTTP port.
  * Without a separate port, metrics remain available through the existing actuator endpoint.
  * Unknown dedicated metrics-server paths now return a clear 404 response.

* **Documentation**
  * Added English and Russian configuration details and YAML examples.

* **Bug Fixes**
  * Invalid Prometheus port values are rejected during startup.


master/src/main/resources/application.yaml
```diff
@@ -35,11 +38,7 @@ management:
     web:
       base-path: /api/actuator
       exposure:
-        include: [ "health" ]
-  prometheus:
-    metrics:
-      export:
-        enabled: ${axelix.master.metrics.prometheus.enabled}
+        include: [ "health", "prometheus" ]
```
mipo256, line 41: We do not need to include the Prometheus endpoint - this endpoint only makes sense if we're using Prometheus on the same port as the main app, and even in this case we declare `PrometheusScrapeEndpoint` manually, to avoid relying on Spring Boot's magic. So, we I think we can (and should) remove it from the list

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/WhitelabelNotFoundHandler.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Default handler for {@link PrometheusMetricsAutoConfiguration}'s {@code prometheusHttpServer}
+ * bean: responds {@code 404} to any path not explicitly registered (e.g. {@code /} or
+ * {@code /-/healthy} once its own handler is disabled), instead of the library's own default
+ * handler, which answers {@code 200} on any unmatched path.
+ *
+ * @author Dmitry Mazurov
+ */
+public class WhitelabelNotFoundHandler implements HttpHandler {
```
mipo256, line 36: There are couple things I want to say.

1. I guess we're supposed to catch some error here or something? Because otherwise it is unclear on why we need `try` at all. Right now this `try` exists just to close the given resource, but that is by itself a bit strange, since this callback does not handle this resource by itself, and it closes this resource that was passed to it from outside.

2. But even more so, the broader question is - do we **_really_** need it **_right now_**? It is often easier to add stuff later than to remove it, so "when in doubt - leave it out". I think we're risking here to end up in a YAGNI situation when we release something, and we're not really sure if anybody will need it, and still - it is a public contract, so we're going to have to support it.

Do we really have a compellign reason to include it? Am I missing something? @dima-bzz

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/ConditionalOnMatchingPrometheusPort.java
```diff
@@ -0,0 +1,48 @@
...
+ * actuator); with {@code matches = false}, activates when they differ (Prometheus exposed
+ * through a dedicated HTTP server).
+ *
+ * @author Dmitry Mazurov
+ */
+@Target({ElementType.TYPE, ElementType.METHOD})
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Conditional(OnMatchingPrometheusPortCondition.class)
+public @interface ConditionalOnMatchingPrometheusPort {
```
mipo256: I understand why the code accepts `boolean` and the rationale behind this move, and I generally am with it.

Still, I think accepting a `boolean` here does not clearly communicate what the conditional actually tries to verify. I think we can make it better by first duplicating the "matching" part:

```suggestion
public @interface ConditionalOnPrometheusPort {
```

And then changing the `boolean` attribute into an `enum` attribute named `value` with values:

- MATCHES_APPLICATION_PORT
- CUSTOM_PORT_CONFIGURED

That would allow us to do stuff like:

`@ConditionalOnPrometheusPort(MATCHES_APPLICATION_PORT)` 

and

`@ConditionalOnPrometheusPort(CUSTOM_PORT_CONFIGURED)`

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnMatchingPrometheusPortCondition.java
```diff
@@ -0,0 +1,80 @@
...
+        boolean expectedMatch = annotation.getBoolean("matches");
+        boolean portsMatch = matchesServerPort(context.getEnvironment());
+
+        if (portsMatch == expectedMatch) {
+            return ConditionOutcome.match();
+        }
+
+        return ConditionOutcome.noMatch(
+                expectedMatch
+                        ? "axelix.master.metrics.prometheus.port does not match server.port, "
```
mipo256: Let's not inline the property names even in condition outcomes - it may mislead users since condition outcomes may be printed out in the logs. We have constants for that.

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnMatchingPrometheusPortConditionTest.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Unit tests for {@link OnMatchingPrometheusPortCondition}.
+ *
+ * @author Dmitry Mazurov
+ */
+class OnMatchingPrometheusPortConditionTest {
+
+    @ParameterizedTest // GH-1520
+    @CsvSource({
```
mipo256: So, there are a couple of things here I want to pinpoint:

1. Although it is possible to use `CsvSource` here, I think it is generally not designed for that. Building arguments in literal strings is just a good developer experience, and it is error-prone. Let's go for `@MethodSource`.

2. This test method is already overaloaded - it tests both the "positive case", when the condition matches, and the "negative case", when the condition is not matched. Let's create two separate tests for that. The necessary setup may happen in local private methods.

3. I understand why mockito is used here. Still, we generally try to avoid mocks as much as possible. Can we try to build an annotation config application context here (the example [can be found here](https://github.com/spring-projects/spring-boot/blob/main/module/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/autoconfigure/OnEnabledDevToolsConditionTests.java#L36))? Or just get by via using `ApplicationContextRunner`?

mipo256 (changes requested):
I like the approach in general. There are a couple of local things I would address, but overall we're pretty good.

Thanks @dima-bzz 

P.S: I have resolved unrelated CodeRabbit comments, but regarding the docs - Code Rabbit AI review agent is right.

master/src/main/resources/application.yaml
dima-bzz: I think it's still needed. Declaring `PrometheusScrapeEndpoint` manually only makes Boot's auto-configuration back off (`@ConditionalOnMissingBean`); the bean is still picked up by `WebEndpointDiscoverer` via `@WebEndpoint(id = "prometheus")` and still goes through the exposure filter, which defaults to `health` only. Drop `prometheus` from `management.endpoints.web.exposure.include` and the endpoint returns 404 — `ActuatorPrometheusEndpointTest` should catch it.

If the concern is that it's pointless when a dedicated port is configured, we could move the exposure out of `application.yaml` and contribute it from the auto-configuration only when the ports match. WDYT?

dima-bzz: I think it's still needed. Declaring `PrometheusScrapeEndpoint` manually only makes Boot's auto-configuration back off (`@ConditionalOnMissingBean`); the bean is still picked up by `WebEndpointDiscoverer` via `@WebEndpoint(id = "prometheus")` and still goes through the exposure filter, which defaults to `health` only. Drop `prometheus` from `management.endpoints.web.exposure.include` and the endpoint returns 404 — `ActuatorPrometheusEndpointTest` should catch it.

If the concern is that it's pointless when a dedicated port is configured, we could move the exposure out of `application.yaml` and contribute it from the auto-configuration only when the ports match. WDYT?

dima-bzz:
Technically CodeRabbit is right, but I’m not sure the information is actually useful to a reader. It’s unlikely anyone would deliberately pass 0 to get a random port, and documenting it feels more likely to confuse than to help. 

WDYT @mipo256

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/WhitelabelNotFoundHandler.java
dima-bzz: The reason it exists: without a handler, any path other than /api/actuator/prometheus on the dedicated server returns [default landing page](https://github.com/prometheus/client_java/blob/main/prometheus-metrics-exporter-httpserver/src/main/java/io/prometheus/metrics/exporter/httpserver/DefaultHandler.java) (the one documenting name[] and debug params). We don’t want Master serving that, so this replaces it with a plain 404 in Spring’s whitelabel style. So it’s fixing what we’d otherwise expose, not a feature added speculatively.

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/ConditionalOnMatchingPrometheusPort.java
dima-bzz: Agreed

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnMatchingPrometheusPortCondition.java
dima-bzz: Agreed

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnMatchingPrometheusPortConditionTest.java
dima-bzz: I will try implementing it

mipo256:
> Technically CodeRabbit is right, but I’m not sure the information is actually useful to a reader. It’s unlikely anyone would deliberately pass 0 to get a random port, and documenting it feels more likely to confuse than to help.

Yeah, I agree. In that case, I would remove the '0' from the possible set of values at all, like, explicitely disallow it. I think allowing them to do that and not documenting it is just a slippery slope so that someone would use it in this way. It may also be a security breach, btw.

master/src/main/resources/application.yaml
mipo256: Yeah, I got it, okay, that makes sense then.

> If the concern is that it's pointless when a dedicated port is configured, we could move the exposure out of application.yaml and contribute it from the auto-configuration only when the ports match. WDYT?

I think this is a bit of overengineering. I would not overcomplicate here.

master/src/main/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/WhitelabelNotFoundHandler.java
mipo256: >  We don’t want Master serving that, so this replaces it with a plain 404 in Spring’s whitelabel style.

But Spring Boot's `PrometheusScrapeEndpoint` by default also returns 200 if the specified metric is not found. I do not understand why we need this handler in our case.

mipo256:
Please, re-request review ones the PR is ready @dima-bzz

dima-bzz:
@mipo256 PR is ready for review

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnPrometheusPortConditionTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // when.
+        contextRunner.run(context -> {
+            // then.
+            assertThat(context).doesNotHaveBean(MatchesApplicationPortMarker.class);
+            assertThat(context).hasSingleBean(CustomPortConfiguredMarker.class);
+        });
+    }
+
+    private static Stream<Arguments> matchingPortCases() {
+        return Stream.of(Arguments.of("8080", null), Arguments.of("8080", "8080"), Arguments.of(null, "8080"));
```
mipo256, line 87: I'd use the `DEFAULT_SERVER_PORT` proeprty instead here to signal the intent of the default port clearly.

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnPrometheusPortConditionTest.java
```diff
@@ -0,0 +1,132 @@
...
+        // when.
+        contextRunner.run(context -> {
+            // then.
+            assertThat(context).hasSingleBean(MatchesApplicationPortMarker.class);
+            assertThat(context).doesNotHaveBean(CustomPortConfiguredMarker.class);
+        });
+    }
+
+    @ParameterizedTest // GH-1520
+    @MethodSource("differingPortCases")
```
mipo256, line 71:
```suggestion
    @MethodSource("differentPortCases")
```

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/OnPrometheusPortConditionTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+    private static Stream<Arguments> matchingPortCases() {
+        return Stream.of(Arguments.of("8080", null), Arguments.of("8080", "8080"), Arguments.of(null, "8080"));
+    }
+
+    private static Stream<Arguments> differingPortCases() {
+        return Stream.of(Arguments.of("8080", String.valueOf(TestPortUtils.findAvailableTcpPortOtherThan(8080))));
+    }
+
+    private static String[] properties(@Nullable String serverPort, @Nullable String prometheusPort) {
```
mipo256, line 94: I think there is not need to create a `private` method like this here. For the sake of tests, we can just inline the key-value pairs just as raw `String` instnaces, right in the body of the paramitrized tests.

master/src/test/java/com/axelixlabs/axelix/master/autoconfiguration/metrics/PrometheusMetricsAutoConfigurationTest.java
```diff
@@ -100,6 +110,80 @@ void shouldWireTheFullPrometheusBeanGraph() {
...
+            contextRunner.run(context -> {
+                // then.
+                assertThat(context).hasSingleBean(PrometheusMeterRegistry.class);
+                assertThat(context).hasSingleBean(HTTPServer.class);
+                assertThat(context).doesNotHaveBean(PrometheusScrapeEndpoint.class);
+            });
+        }
+
+        @Test // GH-1520
+        void shouldFailToStartWhenPortIsZero() {
```
mipo256, line 137: Tests `shouldFailToStartWhenPortIsZero`, `shouldFailToStartWhenPortIsAboveValidRange` and `shouldFailToStartWhenPortIsNegative` can be squashed into a single parameterized test, and also we should not check for the error message. Error message of the exception is not the part of the public contract, so it should not be checked

master/src/test/java/com/axelixlabs/axelix/master/ActuatorPrometheusEndpointTest.java
```diff
@@ -17,94 +17,216 @@
...
+
+import static com.axelixlabs.axelix.master.api.infrastructure.InfrastructureApiPaths.PROMETHEUS_METRICS_SCRAPE_PATH;
+import static com.axelixlabs.axelix.master.autoconfiguration.metrics.PrometheusProperties.PROMETHEUS_PORT_PROPERTY;
+import static com.axelixlabs.axelix.master.autoconfiguration.metrics.PrometheusProperties.SERVER_PORT_PROPERTY;
 import static org.assertj.core.api.Assertions.assertThat;
 
 /**
- * Verifies that the actuator prometheus endpoint on the master is available or absent depending on
- * {@code axelix.master.metrics.prometheus.enabled}, without requiring authentication.
+ * Verifies that Prometheus metrics are exposed differently depending on
```
mipo256, line 43: That javadoc is excessive. We generally do not document the exact behavior of the class/test class, since in this case it risks to diverge from the code inside of it. I'll polish that

master/src/test/java/com/axelixlabs/axelix/master/ActuatorPrometheusEndpointTest.java
```diff
@@ -17,94 +17,216 @@
...
                 "axelix.master.metrics.prometheus.enabled=true",
                 "axelix.master.metrics.prometheus.tags.region=eu-west-1"
             })
-    @AutoConfigureTestRestTemplate
-    class WhenEnabled {
+    class WhenPortDiffersFromServerPort {
+
+        @DynamicPropertySource
+        static void configurePrometheusPort(DynamicPropertyRegistry registry) {
+            registry.add(PROMETHEUS_PORT_PROPERTY, TestSocketUtils::findAvailableTcpPort);
```
mipo256, line 71: That is flaky, let's re-use our own test fixture utlity here that we're about to introduce in this same PR

mipo256 (approved):
The change is pretty good, there is some minor polishing to be done. I will perform it myself later in the master.

Thank you @dima-bzz !

PR #1541, opened 2026-08-25
GH-1540 Roles must grant the authorities they derive from other roles

### LINKED ISSUE #1540: [Feature] Roles must grant the authorities they derive from other roles

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Added effective permissions inherited through nested roles.
  * User permissions and authority metadata now include inherited authorities.
  * Role lookups support complete hierarchies and multiple inheritance paths.
  * Added detection and reporting for circular role relationships.

* **Bug Fixes**
  * Prevented duplicate permissions across multiple inheritance paths.

* **Tests**
  * Added coverage for nested, transitive, and diamond-shaped inheritance, circular relationships, and inherited user permissions.

### DIFF
--- a/common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
+++ b/common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
@@ -37,9 +37,8 @@ public interface Role {
     String getName();
 
     /**
-     * Authorities this role grants on its own, without the ones it reaches through its {@link #getComponents()
-     * components}. Meant for whoever stores or serializes the role as it stands; whoever decides what the holder of
-     * the role may do wants {@link #getEffectiveAuthorities()} instead.
+     * Returns the direct role's authorities, i.e. authorities this role grants on its
+     * own, without the ones that are inherited from other {@link #getComponents()}.
      *
      * @return immutable set of {@link Authority} objects associated with this role
      */
@@ -57,12 +56,8 @@ public interface Role {
     Set<Role> getComponents();
 
     /**
-     * Everything this role grants: its own {@link #getAuthorities() authorities} united with the ones of every role
-     * reachable through its {@link #getComponents() components}, however deep. An authority is not copied into the
-     * deriving role, so granting one to a component shows up here at once.
-     *
-     * <p>No visited set is kept on the way down: a composite is assembled bottom-up out of immutable roles, so the
-     * graph of objects cannot close a cycle whatever the data it was assembled from said.</p>
+     * Everything this role grants: its own {@link #getAuthorities() authorities} combined with the ones of every role
+     * reachable through its {@link #getComponents() components}, however deep.
      *
      * @return immutable set of every {@link Authority} the role grants, directly or through a component
      */
--- a/common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
+++ b/common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
@@ -17,6 +17,7 @@
  */
 package com.axelixlabs.axelix.common.auth.core;
 
+import java.util.HashSet;
 import java.util.Set;
 
 /**
@@ -36,7 +37,9 @@ public interface Role {
     String getName();
 
     /**
-     * Authorities of a given role.
+     * Authorities this role grants on its own, without the ones it reaches through its {@link #getComponents()
+     * components}. Meant for whoever stores or serializes the role as it stands; whoever decides what the holder of
+     * the role may do wants {@link #getEffectiveAuthorities()} instead.
      *
      * @return immutable set of {@link Authority} objects associated with this role
      */
@@ -52,4 +55,24 @@ public interface Role {
      * @return immutable set of {@link Role} objects included in this role
      */
     Set<Role> getComponents();
+
+    /**
+     * Everything this role grants: its own {@link #getAuthorities() authorities} united with the ones of every role
+     * reachable through its {@link #getComponents() components}, however deep. An authority is not copied into the
+     * deriving role, so granting one to a component shows up here at once.
+     *
+     * <p>No visited set is kept on the way down: a composite is assembled bottom-up out of immutable roles, so the
+     * graph of objects cannot close a cycle whatever the data it was assembled from said.</p>
+     *
+     * @return immutable set of every {@link Authority} the role grants, directly or through a component
+     */
+    default Set<Authority> getEffectiveAuthorities() {
+        Set<Authority> all = new HashSet<>(getAuthorities());
+
+        for (Role component : getComponents()) {
+            all.addAll(component.getEffectiveAuthorities());
+        }
+
+        return Set.copyOf(all);
+    }
 }
--- a/common/src/test/java/com/axelixlabs/axelix/common/auth/core/DefaultRoleTest.java
+++ b/common/src/test/java/com/axelixlabs/axelix/common/auth/core/DefaultRoleTest.java
@@ -34,45 +34,45 @@ class DefaultRoleTest {
     @Test
     void shouldReturnItsOwnAuthoritiesWhenItDerivesFromNobody() {
         // given.
-        Role role = new DefaultRole("VIEWER", Set.of(DefaultAuthority.CACHES_TOGGLE));
+        Role role = new DefaultRole("VIEWER", Set.of(OssAuthority.CACHES_TOGGLE));
 
         // when & then.
         assertThat(role.getEffectiveAuthorities())
                 .extracting(Authority::getName)
-                .containsExactly(DefaultAuthority.CACHES_TOGGLE.name());
+                .containsExactly(OssAuthority.CACHES_TOGGLE.name());
     }
 
     @Test
     void shouldUniteTheAuthoritiesOfTheWholeChainItDerivesFrom() {
         // given. C derives from B, B derives from A
-        Role a = new DefaultRole("A", Set.of(DefaultAuthority.ENV_VALUES_READ));
-        Role b = new DefaultRole("B", Set.of(DefaultAuthority.CACHES_CLEAR), Set.of(a));
-        Role c = new DefaultRole("C", Set.of(DefaultAuthority.GARBAGE_COLLECTOR), Set.of(b));
+        Role grandParent = new DefaultRole("A", Set.of(OssAuthority.ENV_VALUES_READ));
+        Role parent = new DefaultRole("B", Set.of(OssAuthority.CACHES_CLEAR), Set.of(grandParent));
+        Role child = new DefaultRole("C", Set.of(OssAuthority.GARBAGE_COLLECTOR), Set.of(parent));
 
         // when & then. The walk goes all the way down rather than one level
-        assertThat(c.getEffectiveAuthorities())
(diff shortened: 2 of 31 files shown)


common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
```diff
@@ -52,4 +55,24 @@ public interface Role {
...
+     * graph of objects cannot close a cycle whatever the data it was assembled from said.</p>
+     *
+     * @return immutable set of every {@link Authority} the role grants, directly or through a component
+     */
+    default Set<Authority> getEffectiveAuthorities() {
+        Set<Authority> all = new HashSet<>(getAuthorities());
+
+        for (Role component : getComponents()) {
+            all.addAll(component.getEffectiveAuthorities());
+        }
```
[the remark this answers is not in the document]
sergeycherkasovv: Skipped. At the moment, this is not a problem that needs to be addressed, especially not in the proposed way. We may revisit it in the future if it becomes an issue.

common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
```diff
@@ -36,7 +37,9 @@ public interface Role {
     String getName();
 
     /**
-     * Authorities of a given role.
+     * Authorities this role grants on its own, without the ones it reaches through its {@link #getComponents()
+     * components}. Meant for whoever stores or serializes the role as it stands; whoever decides what the holder of
+     * the role may do wants {@link #getEffectiveAuthorities()} instead.
```
mipo256:
```suggestion
     * Returns the direct role's authorities, i.e. authorities this role grants on its own, without the ones that are inherited from other {@link #getComponents()
     * components}.
     * 
     * @see getEffectiveAuthorities()
```

common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
```diff
@@ -52,4 +55,24 @@ public interface Role {
      * @return immutable set of {@link Role} objects included in this role
      */
     Set<Role> getComponents();
+
+    /**
+     * Everything this role grants: its own {@link #getAuthorities() authorities} united with the ones of every role
+     * reachable through its {@link #getComponents() components}, however deep. An authority is not copied into the
+     * deriving role, so granting one to a component shows up here at once.
```
mipo256: I do not understand what:

> An authority is not copied into the deriving role, so granting one to a component shows up here at once.

actually means. If it is supposed to communicate something important, let's rephrase; but right now it is unclear. 

```suggestion
     * Everything this role grants: its own {@link #getAuthorities() authorities} united with the ones of every role
     * reachable through its {@link #getComponents() components}, however deep. 
```

common/src/main/java/com/axelixlabs/axelix/common/auth/core/Role.java
```diff
@@ -52,4 +55,24 @@ public interface Role {
...
+     *
+     * <p>No visited set is kept on the way down: a composite is assembled bottom-up out of immutable roles, so the
+     * graph of objects cannot close a cycle whatever the data it was assembled from said.</p>
```
mipo256: Also a pretty obscure sentence 

```suggestion
```

common/src/test/java/com/axelixlabs/axelix/common/auth/core/DefaultRoleTest.java
```diff
@@ -0,0 +1,78 @@
...
+                .extracting(Authority::getName)
+                .containsExactly(DefaultAuthority.CACHES_TOGGLE.name());
+    }
+
+    @Test
+    void shouldUniteTheAuthoritiesOfTheWholeChainItDerivesFrom() {
+        // given. C derives from B, B derives from A
+        Role a = new DefaultRole("A", Set.of(DefaultAuthority.ENV_VALUES_READ));
+        Role b = new DefaultRole("B", Set.of(DefaultAuthority.CACHES_CLEAR), Set.of(a));
+        Role c = new DefaultRole("C", Set.of(DefaultAuthority.GARBAGE_COLLECTOR), Set.of(b));
```
mipo256:
```suggestion
        // given. 
        Role grandParent = new DefaultRole("A", Set.of(DefaultAuthority.ENV_VALUES_READ));
        Role parent = new DefaultRole("B", Set.of(DefaultAuthority.CACHES_CLEAR), Set.of(a));
        Role child = new DefaultRole("C", Set.of(DefaultAuthority.GARBAGE_COLLECTOR), Set.of(b));
```

master/src/test/java/com/axelixlabs/axelix/master/service/auth/DefaultCookieServiceTest.java
```diff
@@ -99,6 +99,29 @@ void buildAuthoritiesMetadataCookie_WithDistinctAuthorities_ReturnsBase64Encoded
         assertThat(decodedValue).contains(DefaultAuthority.ENV_VALUES_READ.name());
     }
 
+    @Test
+    void buildAuthoritiesMetadataCookie_WithInheritedAuthorities_IncludesTheOnesReachedThroughTheComponents() {
+        // given. The front-end decides what to show from this cookie, so an inherited authority has to reach it just
+        // like a direct one
+        Role parent = new DefaultRole("PARENT", Set.of(DefaultAuthority.ENV_VALUES_READ));
```
mipo256: We do not need it really here in this case
```suggestion
```

master/src/test/java/com/axelixlabs/axelix/master/service/auth/DefaultCookieServiceTest.java
```diff
@@ -99,6 +99,29 @@ void buildAuthoritiesMetadataCookie_WithDistinctAuthorities_ReturnsBase64Encoded
...
+        Role grandParent = new DefaultRole("GRAND_PARENT", Set.of(DefaultAuthority.CACHES_CLEAR));
+        Role targetRole = new DefaultRole(
```
mipo256: Let's stick to the "child" naming convention here, since it's used in several places already, and I have proposed to use it in another place as well.

```suggestion
        Role child = new DefaultRole(
```

master/src/test/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleServiceTest.java
```diff
@@ -153,6 +165,148 @@ private static Role roleNamed(Set<Role> roles, String name) {
...
+            jdbcClient
+                    .sql("INSERT INTO roles_authorities (role_id, authority_id)"
+                            + " SELECT ?, id FROM authorities WHERE name = ?")
+                    .params(id, authority.name())
+                    .update();
+
+            return id;
+        }
+
+        private void deriveFrom(String roleId, String parentRoleId) {
```
mipo256: We're using the term "bond" a lot, let's stick with it here as well.

```suggestion
        private void createBond(String childRoleId, String parentRoleId) {
```

master/src/test/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleServiceTest.java
```diff
@@ -153,6 +165,148 @@ private static Role roleNamed(Set<Role> roles, String name) {
...
+                        .containsExactly("CACHES_CLEAR");
+            });
+        }
+
+        @Test
+        void shouldResolveTheBondsTransitively() {
+            // given. C derives from B, B derives from A
+            String roleA = customRole("ROLE A", DefaultAuthority.ENV_VALUES_READ);
+            String roleB = customRole("ROLE B", DefaultAuthority.CACHES_CLEAR);
+            String roleC = customRole("ROLE C", DefaultAuthority.GARBAGE_COLLECTOR);
```
mipo256:
```suggestion
            String grandParent = customRole("GRAND_PARENT", DefaultAuthority.ENV_VALUES_READ);
            String parent = customRole("PARENT", DefaultAuthority.CACHES_CLEAR);
            String child = customRole("CHILD", DefaultAuthority.GARBAGE_COLLECTOR);
```

master/src/test/java/com/axelixlabs/axelix/master/service/state/DefaultRoleServiceTest.java
```diff
@@ -153,6 +165,148 @@ private static Role roleNamed(Set<Role> roles, String name) {
         }
     }
 
+    /**
+     * The bonds of {@code roles_parents} are resolved into the components of the role, which is what makes an
+     * inherited authority reach the checks. The built-in roles carry no bonds, so nothing else in this class covers
+     * it.
+     */
+    @Nested
+    class Inheritance {
```
mipo256, line 179: For all the tests here - I would not use string literals for authorities. That would create a maintenance burden later. Let's stick with just `.name()` on appropriate `enum` instances

master/src/test/java/com/axelixlabs/axelix/master/service/state/DatabaseRoleServiceTest.java
```diff
@@ -153,6 +165,148 @@ private static Role roleNamed(Set<Role> roles, String name) {
...
+            Role role = roleService.findByName("BOTTOM").orElseThrow();
+
+            // then. Reaching the same authority twice is not an error, it is simply unioned
+            assertThat(role.getEffectiveAuthorities())
+                    .extracting(Authority::getName)
+                    .containsExactlyInAnyOrder("GARBAGE_COLLECTOR", "CACHES_CLEAR", "CACHES_TOGGLE", "ENV_VALUES_READ");
+        }
+
+        @Test
+        void shouldTerminateOnABondCycleWrittenAroundTheApplication() {
```
mipo256: I perfectly realize why we would need to make sure it never happens. If for any reason the database state becomes corrupted:

- Malicious user activity
- Our own bug or whatever

We must provide the escape hatch to log in in either case. That is a 100%.

Still, I think that silently parsing cycles is not a good idea. In this case, it is unclear on what the `getComponents` on such roles must return. If we're going to take as the example the current case, when A --> B and B --> A, then in code, once parsed, what this code should evaluate into?

```(java)
var whatGoesHere  = roleA.getComponents()
    .stream()
    .filter(it -> it.getName().equals(roleB.getName()))
    .findFirst().get()
    .getComponents()
```

I'm trying to simulate the case when we're essentially requesting the grandparent of either A or B. In this case, if we return B/A accordingly, **we get cycle**, the exact thing that is going to blow up. But if we do not return anything, or return the incomplete `List`, then in this case the abstraction is only partially right. 

The problem is that in this case the state of the system is corrupted, and we're operating over the corrupted state. I think that this may result a security problem as well.

My proposal is **to throw an exception and fail**, and disslow to create cycles in our domain of the `Role`. And when/if it happens, the end users should use the `SUPER_ADMIN` account, that is going to always be there, and that is going to provide the escape hatch to see what exaclty is the problem and fix it.

mipo256 (approved):
Generally okay, I'm going to take a couple of things to discuss offline.

P.S: upon the merge I'm going to introduce the RoleGraph as the seprate domain abstraction, we're gonna need it.

mipo256 (approved):

PR #1560, opened 2026-08-30
GH-1559 Add the roles feed to the external API

### LINKED ISSUE #1559: [Feature] Add the roles feed to the external API

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Added a protected roles feed endpoint at `/api/external/roles/feed`.
  * Role entries include an identifier, name, description, and number of assigned members.
  * Access requires permission to view users.
  * Member counts default to zero when no users are assigned.

* **Tests**
  * Added coverage for the roles feed endpoint and service behavior, including built-in roles, membership counts, inheritance, and role relationships.

### DIFF
--- a/master/src/main/java/com/axelixlabs/axelix/master/api/external/response/RoleFeedResponse.java
+++ b/master/src/main/java/com/axelixlabs/axelix/master/api/external/response/RoleFeedResponse.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright (C) 2025-2026 Axelix Labs
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package com.axelixlabs.axelix.master.api.external.response;
+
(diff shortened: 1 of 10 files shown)


master/src/main/java/com/axelixlabs/axelix/master/api/external/response/RoleFeedResponse.java
```diff
@@ -15,31 +15,23 @@
...
 
-import org.springframework.test.util.TestSocketUtils;
+import com.axelixlabs.axelix.master.domain.RoleFeed;
 
 /**
- * Test utilities for allocating TCP ports.
+ * Public view of a single role.
  *
- * @author Dmitry Mazurov
+ * @param id           Unique identifier of the role.
```
sergeycherkasovv, line 25: We need this ID so that users can be assigned roles by role ID rather than by role name, since the name is not a reliable identifier. Currently, roles are assigned by name when creating a user. I’m planning to change this in the next PR.

@coderabbitai, please create a follow-up issue to replace the role name with the role ID when creating a user.

master/src/test/java/com/axelixlabs/axelix/master/service/state/DefaultRoleServiceTest.java
```diff
@@ -71,11 +73,26 @@ class DefaultRoleServiceTest {
...
+        jdbcClient.sql("DELETE FROM roles_parents").update();
+        jdbcClient
+                .sql("DELETE FROM roles_authorities WHERE role_id IN (SELECT id FROM roles WHERE role_origin <> ?)")
+                .param(RoleOrigin.BUILT_IN.name())
+                .update();
+        jdbcClient
+                .sql("DELETE FROM roles WHERE role_origin <> ?")
+                .param(RoleOrigin.BUILT_IN.name())
+                .update();
+    }
```
sergeycherkasovv, line 91: I’m not sure this is the right approach to cleaning up the database in tests, but I haven’t found a better way to handle it in this module yet. Let’s keep it for now and revisit it in the future)

master/src/main/java/com/axelixlabs/axelix/master/api/external/response/RoleFeedResponse.java
sergeycherkasovv: @coderabbitai, please create a follow-up issue to replace the role name with the role ID when updating a user.

PR #1624, opened 2026-09-15
Expose Axelix actuator endpoints by default.

### LINKED ISSUE #1592: [Feature] Configure the exposed actuator endpoints for our users automatically
Right now, Axelix as the project, is not ready to work in a condition when only a portion of the endpoints is available. Moreover, this topic is complex, and it is unclear whether we're going to consider that in the nearby future. 

So as of now, in order for the end-user to have a good user experience, in the docs for the starters, we state that all endpoints must be enabled in order for us to work correctly. 

But if we know that all endpoints are needed for Axleix to work, we can configure them ourselves via `EnvironmentPostPorcessor`, and we probably should. The end-user will have less stuff to configure.

guccikray:
The starters now contributes its required actuator endpoints to management.endpoints.web.exposure.include automatically via an EnvironmentPostProcessor, so consumers no longer need to list them manually for Master to reach the service.

Closes #1592

## Summary by CodeRabbit

* **New Features**
  * Axelix management endpoints are exposed automatically by the Spring Boot starter.
  * Custom exposure settings are preserved and combined with the starter’s endpoints.
  * Wildcard exposure (`*`) remains unchanged.
  * Endpoints activate only when their prerequisites are met.
  * Native `health` and `info` endpoints require explicit configuration when needed after overriding defaults.

* **Documentation**
  * Simplified setup instructions by removing manual endpoint exposure configuration.
  * Added an endpoint reference table with descriptions and availability details.


sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/AxelixEndpointsEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+ * Exposes default Axelix endpoints
+ *
+ * @author Marsel Semenov
+ */
+public class AxelixEndpointsEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
+
+    private static final String PROPERTY = "management.endpoints.web.exposure.include";
+    private static final List<String> ENDPOINTS_TO_EXPOSE = List.of(
+            "info",
+            "health",
```
dima-bzz: Let's remove info and health from ENDPOINTS_TO_EXPOSE. The starter should expose only its own axelix-* endpoints. Exposing standard Spring Boot endpoints is the user's decision.

sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/AxelixEndpointsEnvironmentPostProcessor.java
```diff
@@ -0,0 +1,85 @@
...
+            "axelix-metrics",
+            "axelix-scheduled-tasks",
+            "axelix-thread-dump",
+            "axelix-dependencies");
+
+    @Override
+    public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
+        Set<String> current = Binder.get(environment)
+                .bind(PROPERTY, Bindable.setOf(String.class))
+                .orElse(Collections.emptySet());
```
dima-bzz: When the user hasn't set management.endpoints.web.exposure.include, writing our list overrides Spring's default (health). In that case we should keep Spring's default rather than hardcode it.

dima-bzz:
Hey @guccikray! This post-processor is identical across all three starters. Maybe we should move it to `starter-domain` to avoid duplication?

WDYT @mipo256?

mipo256:
> Hey @guccikray! This post-processor is identical across all three starters. Maybe we should move it to starter-domain to avoid duplication?

A very tempting idea, but unfortunately it is not possible. The `starter-domain` module is intentiallly spring-agnostic, and as a result, it cannot carry any code that is specific to Spring Framework. Our `starter-domain` and individual starter modules act as the hexagonal architecture of some sort, where post processor for each version of spring can/may be different, and thus it acts as the adapter, essentailly.

dima-bzz:
Thanks @mipo256 for the clarification!

guccikray:
> where post processor for each version of spring can/may be different

Sure. In Spring Boot 2 and 3, `EnvironmentPostProcessor` is located in `org.springframework.boot.env.` In Spring Boot 4, it has been moved to `org.springframework.boot.`

mipo256:
Thanks, @guccikray, that has been polished and merged. I've preserved your authorship of the patch.

guccikray:
> Thanks, @guccikray, that has been polished and merged. I've preserved your authorship of the patch.

I'm glad to help!

PR #419, opened 2025-12-02
GH-405 Design extensible mechanism for defining invalid application properties

NikitaKirilloff:
closed #405

PR #1629, opened 2026-09-18
GH-1609 Design the dangerous properties API

### LINKED ISSUE #1609: [Feature] Design the dangerous properties API
We need to add the ability to discover the properties that we consider dangerous. What I think we can do is that we can store some kind of dictionary in the `starter-domain` that is going to outline the property key/value combinations that we consider dangerous. I think for now, it is generally safe to say that the given key/value pair is dangerous in all the Spring Boot versions, although, it may change in the future.

What we need to store is:

- The property key as it would be written in the kebab case (which is the spelling case that we use most often in our docs and spring boot [also uses this case](https://docs.spring.io/spring-boot/appendix/application-properties/index.html), I guess...)
- The value that we consider offending/dangerous
- The recommended alternative example

I am not 100% sure about the recommended alternative example. I can easily imagine the case where the given property value is bad, but there is no _"good value"_, rather, there are alternatives that are somewhat better. **I'll leave that decision for the implementer**.  My current stance on this is that we may say that this is just the alternate example that is better, not advised, but an alternative example. We're not saying that this is the value that you should use, but _what can this property value be_.

Another open question is where do we store the rationale description of why we think the given property value is dangerous.

I think that for now we can follow the example of `PropertySourceDescription`, i.e. store it only in the english format. Still, we would have to i18nize someday, but I think for now it is okay.

For now, let's add them only for the response of the actuator endpoints from the starter side. I'm pretty sure that we're going to need it in the heartbeat, but that requires its own design. I will deliberately put that aside for now.

sergeycherkasovv:
## Summary by CodeRabbit

* **New Features**
  * Environment feeds now identify potentially dangerous Spring Boot configuration values.
  * Reports include the reason a value is risky and, where available, a safer alternative example.
  * Detection covers common database, error-handling, SQL logging, lazy-loading, and actuator exposure settings.
  * Dangerous-value details are shown only for the primary occurrence of duplicated properties.
  * Detection continues to provide rationale when property values are masked, while harmless values remain unflagged.

* **Tests**
  * Added coverage for case variations, masked values, duplicate properties, and unresolved placeholders.


sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/DangerousProperty.java
```diff
@@ -0,0 +1,166 @@
...
+        List<DangerousProperty> candidates = BY_NORMALIZED_NAME.get(normalizedPropertyName);
+
+        if (candidates == null) {
+            return null;
+        }
+
+        String trimmedValue = value.trim();
+
+        for (DangerousProperty candidate : candidates) {
+            if (candidate.dangerousValue.equalsIgnoreCase(trimmedValue)) {
```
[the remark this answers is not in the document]
sergeycherkasovv: 1. I think this is valid, but it looks like `on`, `yes`, and `1` are used very rarely - by far the most common value is `true`, which we already handle. We could of course put together a list of values that all mean the same thing, but that's something worth discussing first.

2. I think using a value like `health,*` is a very rare scenario, since it doesn't really make sense - it's equivalent to `*`.

sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/DefaultEnvPropertyEnricher.java
```diff
@@ -152,6 +157,24 @@ private Deprecation buildFromMetadata(@Nullable PropertyMetadata propertyMetadat
...
+
+    @Nullable
+    private String effectiveValue(String propertyName) {
+        try {
+            return environment.getProperty(propertyName);
+        } catch (IllegalArgumentException unresolvablePlaceholder) {
+            return null;
+        }
+    }
+
```
sergeycherkasovv, line 177: I think it's important to explain why this code exists.

Property values are sanitized while Spring builds the `EnvironmentDescriptor` - `EnvironmentEndpoint.describeValueOf` stores the already-masked value - which happens before `enrich(...)` ever sees them. So inside `enrich(...)` the descriptor cannot tell us what the property is really set to, and we ask `Environment` instead.

Why this route:
1. `Environment.getProperty(...)` resolves by source precedence, and we only call it for entries already marked `isPrimary`, so both agree on which source won - we always judge the value the application actually runs with.
2. It requires no changes to the sanitization mechanism.

The alternative would be to move sanitization so that it runs after `enrich(...)`.

sbs/starter-domain/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/DangerousProperty.java
```diff
@@ -0,0 +1,166 @@
...
+
+import org.jspecify.annotations.Nullable;
+
+/**
+ * The dictionary of the property values that we consider dangerous in production. One constant per
+ * {@code (property, value)} pair, so a property with several offending values is listed several times.
+ *
+ * @author Sergey Cherkasov
+ */
+public enum DangerousProperty {
```
sergeycherkasovv, line 33: The [task](https://github.com/axelixlabs/axelix/issues/1609) said we store the descriptions for dangerous properties in an enum. It may also have meant that the property name, the unsafe value, and the alternative value to use should live in the enum as well. Either way, I decided to keep all of it in the enum.

I'm bringing this up because in [this PR](https://github.com/axelixlabs/axelix/pull/419) everything is stored in [common-invalid-properties.yml](https://github.com/axelixlabs/axelix/pull/419/changes#diff-7fcc2ff5b08824e7811050c6019c2aed32718a218c15aac6f1a27c363814510f), and I'm not sure that's the best option, since it requires us to write a service for parsing that YAML file. Unless, of course, we want people to create their own custom YAML files listing their own unsafe property values — but I don't think that parser fits us either, because we have somewhat different plans for how to give people control over their environment.

sbs/axelix-spring-boot-2-starter/src/main/java/com/axelixlabs/axelix/sbs/spring/core/env/DefaultEnvPropertyEnricher.java
mipo256: I understand the motivation here, and I agree with the approach. I'd also document that, since it is not directly obvious.

mipo256 (changes requested):
Following the design principle that we have:

> "Smart Server, dump client"

In which the starter must be as dump as possible, I think, at least the detection of property being dangerous should happen in Axelix Master. In other words, the dictionary, that controls the dangerous properties must reside in the Axelix Master, and Axelix Master must discover whether the property value is dangerous or not. 

That would also allow us later to provide the ability to configure dangerous properties in the enterprise, if such functionality will be needed (and it probably will). We need to re-design that.

@sergeycherkasovv @NikitaKirilloff

NikitaKirilloff:
> In which the starter must be as dump as possible, I think, at least the detection of property being dangerous should happen in Axelix Master. In other words, the dictionary, that controls the dangerous properties must reside in the Axelix Master, and Axelix Master must discover whether the property value is dangerous or not.


@mipo256 , yes, 100% agree

